PR #59, opened 2014-10-30
add debianization

rudenkovk:
We packed all OBS in deb-package for our internal installation . Here is debianization for libsolv.


mlschroe:
AFAIK the debian people don't want to have any packaging stuff in upstream tarballs.
See for example the "What's wrong with upstream shipping a debian/ directory" in  https://wiki.debian.org/DebianMentorsFaq

PR #67, opened 2015-01-22
Added pkg-config file for library.

kirgene:
Signed-off-by: Yevhen Kyriukha <kirgene@gmail.com>


mlschroe:
Hmm, at least SUSE and Fedora put pkgconfig files for 64bit libraries into /usr/lib64/pkgconfig. Is that different for your distro?

kirgene:
On Debian pkg-config doesn't find pkgconfig files in /usr/lib64/pkgconfig so it should be in /usr/lib/pkgconfig.

mlschroe:
Sigh, why must Debian be different? I'll check how other projects do this.

kirgene:
I checked how curl doing it:
https://github.com/bagder/curl/blob/master/CMakeLists.txt#L1136

mlschroe:
But libcurl can be built both with cmake and configure. The latter installs the file in $(libdir)/pkgconfig, which translates to /usr/lib64/pkgconfig.

Would ${LIB_INSTALL_DIR}/pkgconfig also work for debian? That's probably "/usr/lib/x86_64-linux-gnu/pkgconfig" on your system.

kirgene:
${LIB_INSTALL_DIR}/pkgconfig is /usr/lib64/pkgconfig on my machine. Tested on Ubuntu 14.10 and Debian 7.

mlschroe:
So this is Debian before the multi-arch change? But shouldn't pkg-config also look in that dir? How do you do arch specific pkgconfig files if it only looks in /usr/lib/pkgconfig?

Anyway, I checked a couple of projects and most of them use ${LIB_INSTALL_DIR}/pkgconfig. So I'm tempted to use this and allow to override it with some cmake option.

kirgene:
On Ubuntu x64 I have `/usr/lib` and `/usr/lib32`, also `/usr/lib/i386-linux-gnu` and `/usr/lib/x86_64-linux-gnu` are present. But cmake expands `${LIB_INSTALL_DIR}` to `/usr/lib64`.

mlschroe:
But that's a bug in libsolv's CMakeFile, it should use /usr/lib/i386-linux-gnu on Debian...

(The libsolv package in Jessie calls cmake with ```-DLIB="lib/$(shell dpkg-architecture -qDEB_BUILD_MULTIARCH)"``` as a workaround...)

kirgene:
OK, I'll update pull request to use ${LIB_INSTALL_DIR}/pkgconfig.
But I think that this issue needs to be fixed also. I don't quite understand is this a bug in CMakeFile or in cmake itself?

cgwalters:
pkg-config looks in `/usr/share/pkg-config` for architecture independent packages.  Using `$libdir` in automake parlance for arch-dependent `.pc` files is right.  I believe cmake's `${LIB_INSTALL_DIR}` is equivalent.

EricAustinYu:
I stumbled upon this pull request while searching around as to whether libsolv had a pkg-config file. It seems this request was made a while ago. Is there any news on whether this functionality will be officially added to libsolv?

mlschroe:
Thanks for the reminder. I'll try to add it.

PR #68, opened 2015-01-29
Save "Status" field of debian packages during parsing of packages list.

kirgene:
When populating packages from "/var/lib/dpkg/status" into repo that should be marked as "installed repo",
there should be possible to filter partially installed packages based on their "Status" field.
This commit makes saving this field for later manual filtering.


mlschroe:
I renamed it to SOLVABLE_INSTALLSTATUS, maybe it is also useful for other package types.

PR #75, opened 2015-02-04
Introduced new SOLVABLE_USERINSTALLED attribute to mark user-installed p...

kirgene:
...ackages during cleandeps process.


mlschroe:
But... but... is there really a "auto-installed" line in the dpkg status file? I thought they live only in /var/lib/apt/extended_states or maybe /var/lib/aptitude/pkgstates.

kirgene:
Ah, I forgot to mention that there are some dpkg compatible package managers like opkg (used in embedded systems). These package managers store all status info in status file.

ignatenkobrain:
ping @kirgene

ignatenkobrain:
Seems nobody is interested in this. Please open a new PR in case this is still relevant.

PR #76, opened 2015-02-11
Better handling of DEBIAN package format.

kirgene:


ignatenkobrain:
This need rebase and I wonder if it's still needed. Also better commit message would be nice.

kirgene:
This pool request is not actual anymore

PR #83, opened 2015-05-08
evr.c cleanup

bmwiedemann:
This change makes code more readable by using ctype-style macros to reduce code-duplication.
I tested that it still compiles and output from gcc -E src/evr.c looks good.

### DIFF
--- a/src/ctype2.h
+++ b/src/ctype2.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2011, Novell Inc.
+ *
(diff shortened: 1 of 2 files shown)


mlschroe:
Hmm, I don't think it improved readability that much.

bmwiedemann:
Looking especially at the alpha-numeric case and the negated variant thereof, I think it makes a big difference there.

mlschroe:
Don't what it. Sorry.

PR #90, opened 2015-06-10
Implement Tcl bindings

neverpanic:
This set of changes
 - Implements (experimental) Tcl bindings for libsolv
 - Adds a few #warning pre-processor macros to simplify porting the bindings to new languages
 - Fixes the libsolv library install names on OS X
 - Fixes a bug detected by clang.

No extensive testing of the Tcl bindings has been done yet, but we will fix problems as we encounter them.


mlschroe:
Cool, thanks!

mlschroe:
I guess this is for the MacPorts GSoC project?

neverpanic:
Correct. https://trac.macports.org/wiki/SummerOfCode2015#dependencies has some information about it.

We initially looked into a CUDF-based solver, but libsolv seems to be faster, more widely tested and deployed and more stable. Of course MacPorts has a couple of rough spots and requirements that may not really fit into what libsolv expects from its solvables, so let's see how this will work out in the end :)

PR #93, opened 2015-06-29
Ship patcheck and avoid false positive errors with src rpms

lchiquitto:
Newer patch files include conflicts on src/nosrc packages. Example:

$ zypper info -t patch  SUSE-SLE-WE-12-2015-59
(...)
Provides:
  patch:SUSE-SLE-WE-12-2015-59 == 1
Conflicts:
  srcpackage:libvirt < 1.2.5-21.1
  libvirt-client-32bit.x86_64 < 1.2.5-21.1

In practice though, such conflicts can never prevent the installation of a patch, as installed src/nosrc packages are not recorded in the RPM database.

This fix makes patchcheck filter out src/nosrc RPMs from its "to install" repository, preventing it
finding lots of conflicts that won't happen in the real world.

### DIFF
--- a/tools/patchcheck.c
+++ b/tools/patchcheck.c
@@ -158,6 +158,9 @@ toinst(Solver *solv, Repo *repo, Repo *instrepo)
       if (p < 0 || p == SYSTEMSOLVABLE)
 	continue;
 
+      Solvable *s = pool->solvables + p;
+      if (s->arch == ARCH_SRC || s->arch == ARCH_NOSRC)
+        continue;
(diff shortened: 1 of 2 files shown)


tools/patchcheck.c
```diff
@@ -158,6 +158,9 @@ toinst(Solver *solv, Repo *repo, Repo *instrepo)
       if (p < 0 || p == SYSTEMSOLVABLE)
 	continue;
 
+      Solvable *s = pool->solvables + p;
```
ignatenkobrain: I would use pool_id2solvable() here

PR #97, opened 2015-08-04
python: drop wrong set of PythonInterp_FIND_VERSION

ignatenkobrain:
Now upstream CMake taking care about consistent versions between
libs and interpreter.


mlschroe:
It should be fixed with commit 204406df09b45e1316f02f1f629ef79574530b3d.

PR #102, opened 2015-09-03
Create autoconf build for Tcl bindings.

JacksonIsaac:
Create autoconf build.

Currently added support for Tcl bindings. This removes the 
added dependency of installing cmake on the system. It also
makes it easy to build it during MacPorts installation as it can be
added as a subproject to the autoconf build.

Assigned the bug report id to macports-dev ML as we were working on it
and happy to maintain it in future.

Please feel free to update the id to upstream ML :).


mlschroe:
Thanks, but I prefer to stay with cmake for now.

PR #105, opened 2015-10-02
puts $op instead of $p in Transaction summary.

JacksonIsaac:
Print $op in the second part of puts when the package is to be
upgraded or downgraded. Previously same package info was been
printed.


mlschroe:
Thanks!

PR #106, opened 2015-10-02
Turn on hawkey-specific behaviors in libsolv for Mageia for DNF

Conan-Kudo:
As part of the work to [enable DNF for Mageia as an alternate repository manager](https://wiki.mageia.org/en/Feature:Add_DNF_as_Alternate_Repository_Manager), definitions expected by hawkey from libsolv need to be set for the MAGEIA definition as well.


Conan-Kudo:
As of today, [DNF has been approved as a secondary package/repository manager for Mageia 6](http://meetbot.mageia.org/mageia-dev/2015/mageia-dev.2015-11-11-20.09.html). 

I would appreciate it if this pull request was reviewed and merged as soon as possible.

mlschroe:
(back from parental leave)

Hmm, I don't see the need.

The first commit changes implicitobsoleteusescolors and adds some legacy architecture. The implicitobsoleteusescolors flag is something that describes how rpm works, it has nothing to do with the package manager. It describes if rpm allows to install a 32bit package *of the same name* next to a 64bit package. This is true for Fedora, but false for SUSE. Does Mageia have multilib support?

About the legacy architectures, they were added to be compatible with some obsolete RHEL version, I don't see how this is relevant for Mageia.

The second commit can also be done by calling cmake with -DENABLE_RPMMD=1. Why don't you add that in the Mageia specfile?

Conan-Kudo:
The first commit were all changes I needed to do to get hawkey tests to stop failing.

That said, Mageia does implement multilib, but the packages aren't the same name yet because the legacy package manager (urpmi) does not support it. However, that could change in the future (if/when urpmi is dropped). Architecturally speaking, Mageia does have all the underpinnings necessary to do so, and we do allow side-by-side 32-bit/64-bit packages to be installed.

Currently, the limitations of urpmi force us to label packages installed to `/usr/lib64` (which is basically all 64-bit libraries) as `lib64<libraryname>` and 32-bit ones that install to `/usr/lib` as `lib<libraryname>`. But for all practical purposes, there's nothing that prevents us from supporting multilib as Fedora uses it with DNF.

The second commit was requested so that it would always be defined whenever someone wants to make libsolv builds targeting Mageia.

PR #113, opened 2015-11-13
application: Don't use an appdata() requires if the package name is k…

mlandres:
…nown

The need for guessing the `appdata(SOME_UNIQ_STRING)` provides, which may or may not have been inserted into a package providing application data, could be avoided by using the <pkgname> hint, if provided by the application metadata.

Todo: For installed application one could lookup the package that installed the appdata.xml rather than guessing it's provides. With this, extra appdata-provides in packages  could be superfluous.

### DIFF
--- a/ext/repo_appdata.c
+++ b/ext/repo_appdata.c
@@ -107,6 +107,7 @@ struct parsedata {
   int flags;
   char *desktop_file;
   int havesummary;
+  Id require;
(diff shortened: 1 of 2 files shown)


mlandres:
Obsoleted by 24e0c4f8acae65006dda641aba82a7916e17c03a ...

PR #123, opened 2016-02-08
Support getting headers from rpmdb via librpm

ffesti:
instead of opening the database directly.

This comes with some speed penalty as the header is loaded first and then reassembled to binary form. I am also working on the second step to actually use librpm to access the header which should get most of this speed back.

### DIFF
--- a/ext/repo_rpmdb.c
+++ b/ext/repo_rpmdb.c
@@ -39,6 +39,11 @@
 # endif
 #endif
 
+#ifdef ENABLE_RPMDB_BYRPMHEADER
+#include <rpm/rpmts.h>
+#include <rpm/rpmmacro.h>
+#endif
+
 #include "pool.h"
 #include "repo.h"
 #include "hash.h"
@@ -1053,10 +1058,13 @@ struct rpmdbstate {
 
   RpmHead *rpmhead;	/* header storage space */
   int rpmheadsize;
-
+#ifdef ENABLE_RPMDB_BYRPMHEADER
+  rpmts ts;
+#else
   int dbopened;
   DB_ENV *dbenv;	/* database environment */
   DB *db;		/* packages database */
+#endif
   int byteswapped;	/* endianess of packages database */
   int is_ostree;	/* read-only db that lives in /usr/share/rpm */
 };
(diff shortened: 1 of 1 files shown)


jsilhan:
FYI @j-mracek tested this with DNF. First DNF transaction succeeds, then it hangs on first item of RPM transaction. Probably it does not close rpmlib correctly when fetching the data.

@ffesti , this should work with BDB and newrpmdb rpm backend, right?

j-mracek:
The transaction freezes nearly always appear with install or remove of acpi package. I would recommend to used acpi package for a testing cases.

j-mracek:
The libsolv version that includes this pullrequest can be downloaded from this location: http://koji.fedoraproject.org/koji/taskinfo?taskID=13373978

mlschroe:
(I just glanced at the commit, so this might be wrong. But isn't there some code missing that calls rpmtsFree and thus closes the db again?)

ffesti:
Yup. Looks like this line didn't make it when refactoring out this commit. Added.

jsilhan:
@j-mracek, retest this please :)

j-mracek:
@ffesti @jsilhan Unfortunately same problem. Here are the new rpms: http://koji.fedoraproject.org/koji/taskinfo?taskID=13381625

ignatenkobrain:
any news here? Should I test it?

ignatenkobrain:
@ffesti could you rebase this PR?

Conan-Kudo:
@ffesti Could you _please_ rebase this PR? It may fix #227.

pmatilai:
@ffesti ping, this is getting more and more important now that we have multiple database backends in rpm. It'd be nice to be able to dogfood those alternative dbs but that's not possible with dnf atm, because of this.

pmatilai:
Here's a rebased version with a bunch of other necessary fixes (explained in commit message):
http://laiskiainen.org/tmp/0001-Support-getting-headers-from-rpmdb-via-librpm-1.patch

There are certainly rough edges there still etc, but with this patch + a small kludge to libdnf, the basic smoketest of dnf install to a chroot and querying the result with an alternative rpmdb backend configured passes.

As for performance, on my laptop's with an rpmdb of ~3300 packages, rpmdb2solv takes 0.78s with the libsolv direct BDB implementation, 1.02s with the librpm patch. So it's not exactly a huge difference, compared to not working at all. 

I also did a rough-cut version that uses librpm's header code to eliminate overhead from conversions, that gets it down to 0.90s for the same data. It's not entirely functional so take that with a grain of salt for now.

pmatilai:
And now that I got it working for the same dnf smoketest, here's the other version that adds support for using librpm code for both rpmdb and header access, gaining back about half of the lost time (ie that 0.90s version):
http://laiskiainen.org/tmp/0001-Support-using-librpm-for-header-and-database-access-1.patch

I'm not suggesting either of these for merging as-is, this is just for testing and discussion purposes.

The amount of code duplication in the header-and-database version is intolerable, but there's quite a bit of that in the earlier version too, and the earlier version gets really hard to read with the #ifdef's. There are a number of ways to reduce the duplication and clean it all up, but before putting more effort into it I'd like some feedback on the direction:

@mlschroe , to what extent you want the local libsolv implementations to be preserved as alternatives? I can kinda understand not wanting to link to librpm on non-rpm distros but those wouldn't need rpmdb support I think. So one possible cutting line could be always using librpm code for headers + rpmdb access if rpmdb support is built in, and leave the libsolv header implementation there for non-native rpm distros but drop the local BDB reading code. Of course the BDB code can be preserved there too, but there's more work to refactor + abstract then.

Conan-Kudo:
@mlschroe Ping? Could you please respond to @pmatilai's query?

Conan-Kudo:
@mlschroe ping?

Conan-Kudo:
@mlschroe pingy ping?

Conan-Kudo:
@mlschroe 

Could you please take a look at @pmatilai's proposed patch (either in the link in his comment, or look at #235 for the PR version) and respond to his comment?

There's now at least two independent efforts to try to use RPM with a non-BDB database (this one and #251).

In addition, my understanding is that generally speaking, things should not be accessing the rpmdb directly as librpm is really the only "stable interface".

Without a fix for this, Fedora [can't change the rpmdb format](https://fedoraproject.org/wiki/Changes/NewRpmDBFormat), and Photon had to chainsaw in support for LMDB to switch to that.

We want to move away from BDB to _something_ supported, and we can't while we don't know how to proceed.

mlschroe:
I don't have much spare time for libsolv time at the moment. I will be busy next week, but I'll try to review/merge after that.

mlschroe:
Please give commit ae214a654c4c0f3b43ca763e70bde4793d00422a a try!

pmatilai:
Awesome, thanks @mlschroe ! Running it on my system now (including the couple of later fixes) and basics seem okay. 

Now to hunt down the other places that make assumptions about rpmdb format/name, for example libdnf (https://github.com/rpm-software-management/libdnf/issues/362)

ignatenkobrain:
@pmatilai unfortunately even now, libsolv has to guess rpmdb name as well (in order to check whether something changed or not). So there is still some problem in libsolv.

DNF/libdnf part should not be a problem since it doesn't cache it.

PR #133, opened 2016-04-21
Define architectures for debian based distros

dburm:
Add debian architecture list


mlschroe:
I'm doing this a bit different with the latest commit. If the architecture is not known, libsolv will no longer use "" as arch policy, but allow the specified arch. This means, the archpolicies array no longer needs to include arch -> arch mappings, so your commit is no longer necessary.

I'm currently not sure if I should make that archpolicies array disttype specific, as I don't think it is needed for debian (and maybe it is even harmful).

PR #134, opened 2016-05-16
CMakeLists.txt: fix MAN_INSTALL_DIR

maxinbjohn:
It checks ${CMAKE_INSTALL_PREFIX}/share/man when configure, but it may
not exist when crosscompile, for example, when CMAKE_INSTALL_PREFIX="/",
it is OK, but when CMAKE_INSTALL_PREFIX="/some/path/else", then it
doesn't exist, and the man dir would be set to "/usr/man" which is
incorrect.

Signed-off-by: Robert Yang <liezhi.yang@windriver.com>
Signed-off-by: Maxin B. John <maxin.john@intel.com>


mlschroe:
Hmm, how about turning the test around so that it reads:
```
maninstalldir=prefix/share/man
if exists prefix/man and not exists prefix/share/man
  maninstalldir=prefix/man
fi
```

would that work for you as well?

maxinbjohn:
Thanks.  Yes, that works fine for me.  I have updated the commit and force pushed it.

PR #144, opened 2016-07-19
Enable libsolv shared library, enable bz2/xz compression support

Conan-Kudo:
When @M0ses and I started working on building libhif on openSUSE, we encountered issues linking libsolv with libhif because libxml2 symbols weren't resolving, and we were missing bzip2 and xz compression support.

This pull request resolves these issues.

### DIFF
--- a/package/libsolv.changes
+++ b/package/libsolv.changes
@@ -1,4 +1,15 @@
 -------------------------------------------------------------------
+Tue Jul 19 09:28:15 UTC 2016 - ngompa13@gmail.com
+
+- Enable libsolv shared library package to fix linking issues
+  with libxml2 and compression libraries
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mlschroe:
Why would we want libhif in openSUSE?

mlschroe:
There must be a better way to fix the linking issues. We use static linking of libsolv on purpose.

Conan-Kudo:
@mlschroe We want to use it for the rpm engine for snapcraft so that there's a common engine for all major RPM based distributions.

Why do you use static linking of libsolv? That said, this PR doesn't disable the static library.

In any case, I would imagine that static libsolv would need static libxml2, zlib, bzip2, and xz linked into its archive, rather than whatever it's doing now.

Conan-Kudo:
Closed in favor of #195.

PR #148, opened 2016-07-20
Also scan /usr/share/metainfo for appdata files

badshah400:
According to the updated [appstream specification](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html) :

>Applications are a special case here, because they are usually treated differently by software centers (and also for historical reasons). If your metainfo file covers only an application, as described in the AppData section, install it as /usr/share/metainfo/%{id}.appdata.xml.
>
>The /usr/share/appdata/ path must be scanned by AppStream tools as well, to support legacy applications installing metadata there. For new software components, it is advised not to use this directory. 

This PR ensures libsolv also scans the /usr/share/metainfo directory for appstream files in addition to the (now legacy) dir /usr/share/appdata supported currently.

### DIFF
--- a/examples/solv/repoinfo_system_rpm.c
+++ b/examples/solv/repoinfo_system_rpm.c
@@ -27,7 +27,8 @@
 # define PRODUCTS_PATH "/etc/products.d"
 #endif
 #ifdef ENABLE_APPDATA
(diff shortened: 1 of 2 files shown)


badshah400:
See also https://bugzilla.opensuse.org/show_bug.cgi?id=989830

PR #164, opened 2016-11-28
Fbsd compat

roman-neuhauser:
*   let users define PKGCONFIG_INSTALL_DIR
    
    FreeBSD wants *.pc files in ${PREFIX}/libdata/pkgconfig,
    ${PREFIX}/lib/pkgconfig is not good eanough a configuration point.

*   let users override MAN_INSTALL_DIR
    
    this makes the build more cooperative with FreeBSD ports:
    FreeBSD wants man pages in ${PREFIX}/man, yet ${PREFIX}/share/man
    exists in the staging area as well for some reason.  i'm too lazy
    to chase the cause, esp. as the copout means more freedom to users.


mlschroe:
Thanks!

PR #186, opened 2017-03-10
installcheck: Fix noarch id setting

a-evseev:
pool_setarch calls also pool_setarchpolicy which uses value of noarchid,
but at the time when pool_setarch is called, noarchid is set to default
value, and this leads to unresolvables for repos, other than RPM (using
`all` or `any` instead of `noarch`).

I'm not sure if pool arch must be set before adding repos, so this is not very clean solution - pool_setdisttype and pool_setarch will be called for every loaded metadata file.

### DIFF
--- a/tools/installcheck.c
+++ b/tools/installcheck.c
@@ -93,8 +93,6 @@ main(int argc, char **argv)
 
   arch = argv[1];
   pool = pool_create();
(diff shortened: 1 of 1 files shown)


a-evseev:
Looks like this breaks installcheck completely :-(

PR #196, opened 2017-05-09
Fix reading 64bit integer values from RPM header

jrohel:
I tried fix DNF (https://bugzilla.redhat.com/show_bug.cgi?id=1300801) and found that bug is in libsolv.
The result of << and | must be cast to "unsigned". Or we get really bad result for some values.

Btw: Why libsolv reimplement librpm instead of using it?


Conan-Kudo:
@jrohel Presumably, it was done for speed. @ffesti gave a shot at moving it to use librpm in #123, and that's the proper solution, really. If it doesn't, then it's impossible to use any libsolv based resolver with RPM's non-BDB based rpmdb backends.

mlschroe:
It's because tools like OBS just need to read rpm headers and there's no need to pull in all libraries needed by librpm for that.

mlschroe:
Thank you for hunting that one down!

PR #209, opened 2017-05-25
_GNU_SOURCE should be defined before including any headers

jon-turney:
per https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html
"These directives must come before any #include of a system header file."

This actually causes a problem when compiling with newlib headers.


mlschroe:
Thanks!

PR #141, opened 2016-07-01
build: use GNUInstallDirs

ignatenkobrain:
Closes: #128
Signed-off-by: Igor Gnatenko ignatenko@redhat.com

PR #211, opened 2017-06-05
Fixes for building with newlib/cygwin

jon-turney:


mlschroe:
I seem to remember that the cmake folks consider it bad style to include the prefix with the DESTINATION. Would `RUNTIME DESTINATION bin` also work?

ignatenkobrain:
@mlschroe it will, because it is completely same (but still I would rather do #141)

jon-turney:
> I seem to remember that the cmake folks consider it bad style to include the prefix with the DESTINATION. Would RUNTIME DESTINATION bin also work?

Makes sense. Done.

PR #219, opened 2017-08-06
Add missing definitions to libsolv for Mageia

Conan-Kudo:
This PR adds the missing definitions to libsolv to make it behave as expected on Mageia (that is, like it does on Fedora). This includes adjustments to the example `solv` tool as well as enabling codepaths previously guarded out for only Fedora, despite also applying to Mageia now.


Conan-Kudo:
@mlschroe Ping?

I've been shipping these fixes in the Mageia package for libsolv, and IMO, it makes sense to have these merged into mainline libsolv.

Conan-Kudo:
Thanks, @mlschroe!

PR #221, opened 2017-08-22
policy.c: Fix solver_prune_to_highest_prio_per_name

EricAustinYu:
- Fixed logic so that all packages with the same name are pushed to the queue
  of packages to be pruned instead of just the first package with that name.
- This fixes a bug where the best version of recommended and suggested packages
  were not being returned.
- Added tests to check that this does fix the problem with recommended and
  suggested packages mentioned above.


ignatenkobrain:
test cases look good to me, not sure what was behavior before though ;)

PR #230, opened 2017-09-30
repo_rpmdb.c: increase MAX_HDR_CNT and MAX_HDR_DSIZE

liuming50:
We encountered 'corrupt rpm' issues when installing extreme big RPM
packages like the kernel-devsrc package of Yocto project.

It can be fixed by increasing MAX_HDR_CNT and MAX_HDR_DSIZE per test.

Signed-off-by: Ming Liu <liu.ming50@gmail.com>

### DIFF
--- a/ext/repo_rpmdb.c
+++ b/ext/repo_rpmdb.c
@@ -170,8 +170,8 @@
 #define MAX_SIG_CNT		0x100000
 #define MAX_SIG_DSIZE		0x100000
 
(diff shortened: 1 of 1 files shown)


liuming50:
May I know how could we sync these 2 macros with RPM project? I can ping RPM community if you can kindly show me how rpm defines these 2 values since I just could not find them in RPM source.

PR #237, opened 2017-11-16
[WIP] Format using clang-format

ignatenkobrain:
libsolv is mixture of tabs/spaces, inconsistent indentation and so on. It would be really nice to get into one code style.

Note that I am tryin to keep original style as much as possible, so still WIP. Especially header files look different yet.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,23 @@
+---
+Language: Cpp
+BasedOnStyle: LLVM
+BreakBeforeBraces: Custom # GNU
+BraceWrapping:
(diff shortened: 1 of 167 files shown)


Conan-Kudo:
Nice! Though if this is merged, it'd be really helpful to have a release right after this, because backports are going to be murderously awful otherwise.

ignatenkobrain:
@Conan-Kudo yeah, this would be good for 1.0 I think

ignatenkobrain:
@mlschroe any opinion on this?

PR #247, opened 2017-12-10
bindings: allow to change repo for solvable

ignatenkobrain:


mlschroe:
That won't work if you have dependencies...

PR #249, opened 2017-12-20
add .cz to ENABLE_ZLIB_COMPRESSION case

termim:
In Mandrake/Mageia/Rosa synthesis.hdlist.cz is compressed using ZLIB.
This fix makes it possible to use xfopen_fd to transparently read it.


Conan-Kudo:
@termim Actually, both Mageia and OpenMandriva use XZ compression for the synthesis hdlists file.

termim:
You are right, it is just Rosa that uses gzip for the synthesis.hdlist.cz file.
I should've verified first, sorry.

PR #235, opened 2017-11-02
WIP: Support using librpm for header and database access (from @pmatilai and @ffesti)

Conan-Kudo:
This PR supersedes #123 and includes the reworked version of @ffesti's patch from @pmatilai.

PR #251, opened 2018-01-10
add lmdb support for rpm

ppadmavilasom:
submitting this PR for discussion.

built with -DPHOTON=ON

tested with the following setup
rpm - version 4.14.0 with experimental support for lmdb
converted existing rpm db in /var/lib/rpm to lmdb using the following export and import scripts
https://github.com/rpm-software-management/rpm/files/1180126/bdbexport.sh.gz
https://github.com/rpm-software-management/rpm/files/1180129/lmdbimport.sh.gz

/var/lib/rpm contains the following 
-rw-r--r-- 1 root root 268435456 2018-01-10 22:30 data.mdb
-rw-r--r-- 1 root root      8192 2018-01-10 22:31 lock.mdb

I was testing on photon so i tested with tdnf using libsolv. Please look it over and see if this is something you will consider for lmdb support.
Thanks!


ignatenkobrain:
I wonder why not to finish with making libsolv to use librpm....

ignatenkobrain:
especially because it should depend on %_db_backend or something like this in RPM and be runtime configurable in RPM.

ppadmavilasom:
just wanted to propose something that kept the libsolv code as is as much as possible.

ignatenkobrain:
see #235

ppadmavilasom:
@ignatenkobrain i checked the PR you pointed out
My reasons for this PR as follows
rpm supports runtime switch of backend db but as far as a distro is concerned, this might not be valuable as there are size/speed concerns which distros make a tradeoff on and stick with it.
libsolv appeared to take a minimum dependency approach (when possible) for external components so considering a new rpm db (whatever that is) libsolv might be just interested in the interface that it requires and nothing more.
keeping legacy libsolv code as is and provide an adaptor approach seemed valuable as the robust path forward.

Conan-Kudo:
@ppadmavilasom This is not necessary as of libsolv 0.6.31.

mlschroe:
Didn't you read his comment for the rationale?

(But with the latest changes in libsolv the pull request should be redone to match what we now have in repo_rpmdb_bdb.c)

Conan-Kudo:
@mlschroe The other thing is that the LMDB structure is _experimental_ and subject to change, principally because there are current shortcomings in LMDB that rpm will need to figure out a way to overcome if it's not fixed in LMDB itself. (c.f. [rh#1086784#c21](https://bugzilla.redhat.com/show_bug.cgi?id=1086784#c21)). I'm not sure it's a good idea to rely on the LMDB data format while you can't even guarantee that it _may_ be somewhat stable.

But if @ppadmavilasom wants to continue down this path, then yes, it should be done as `repo_rpmdb_lmdb.c` with the correct options. At that point, also the RPMDB options should be refactored to account for multiple DB formats...

ppadmavilasom:
@mlschroe thanks for considering. i will resolve conflicts, fill some gaps and update.
@Conan-Kudo appreciate the comments and pointer about lmdb stability w.r.t usage in rpm.

ignatenkobrain:
@ppadmavilasom mind rebasing?

Conan-Kudo:
@mlschroe This can be closed, since LMDB has been removed from RPM entirely.

PR #257, opened 2018-02-26
bindings: use $self for Selection.clone()

ignatenkobrain:

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -1298,12 +1298,12 @@ typedef struct {
     return $self->q.count == 0;
   }
   %newobject clone;
-  Selection *clone(Selection *from, int flags = 0) { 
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Doesn't this break API incompatibly?

ignatenkobrain:
@Conan-Kudo yes, except that I wonder if anyone used this API because it's so wrong... `sel.clone(sel)`

mlschroe:
Hmm, yes. Seems like a bad copy/n/paste from my side. It's also not documented. I think it got stuck in mid-implementation. ;)

ignatenkobrain:
alright, I will add documentation.

mlschroe:
You're supposed to patch the .txt file, not the generated .3 file...

PR #259, opened 2018-03-22
Fix #235 : try to detect data.mdb for lmdb also

vitan:


vitan:
@Conan-Kudo  can you detail the broken logic plz? I tested it under BDB, seems it works still

Conan-Kudo:
@vitan Your logic only works as long as a `data.mdb` file exists. If you start with BDB and don't have an LMDB rpmdb on disk, that logic never evaluates.

Conan-Kudo:
This is no longer needed, as it has been fixed differently in ea037722b0f2a1afc8416e08b36fd498a8bb14c4.

This can be closed.

Conan-Kudo:
@mlschroe This can be closed, since LMDB has been removed from RPM entirely.

vitan:
> @mlschroe This can be closed, since LMDB has been removed from RPM entirely.

hi, how are you. where I can get the reason that lmdb removed ?

Conan-Kudo:
The reason is declared in https://github.com/rpm-software-management/rpm/commit/7de982ac0957c42f228b43685d9f486e55eac331

mlschroe:
Closing as we're following rpm upstream in this regard.

PR #268, opened 2018-05-24
Not remove gpg-pubkey from rpmdb (RhBug:1577170)

j-mracek:
It will allow to remove gpg-pubkey from system, because it will be handled as
a package.

https://bugzilla.redhat.com/show_bug.cgi?id=1577170


Conan-Kudo:
Maybe I'm talking out of left field here, but wouldn't it make sense to export gpg-pubkey "packages" as a special thing so they can easily identified and managed from any libsolv-using package manager?

mlschroe:
I think we'll need to make this an option so that DNF can have gpg-pubkey but SUSE will strip them.

mlschroe:
(The difference is that SUSE still has them in the rpmdb, whereas Fedora has the pubkeys in /var/lib/rpm/pubkeys)

mlschroe:
(Or not? Anyway, they are hidden from package management in SUSE, pubkey management is a different module in YAST.)

Conan-Kudo:
The pubkeys are definitely in the same place in Fedora and openSUSE (at least in Leap 15.0).

But I think it definitely warrants being exposed as a new type of solvable thing (like how patterns, comps, appdata, etc. are handled). That way, it's easy for libzypp to ignore it, while libdnf can do something with it.

mlschroe:
It's not that easy, the solver needs to do deal with them a bit different, i.e. like with multiversion packages.

(And libsolv doesn't really have package types, there are just some conventions about package name prefixes.)

j-mracek:
I create also PR for libdnf to set gpg-pubkey as a SOLVER_MULTIVERSION. I am thinking how to make it optional. I believe that so far we do not have a suitable build flag. Or am I wrong?

mlschroe:
"Optional" means a new RPM_KEEP_GPG_PUBKEY flag in repo_rpmdb.h.

j-mracek:
Sorry I named it as ENABLE_RPMDB_GPG_PUBKEY_REMOVAL, but I can change it.

mlschroe:
`ENABLE_RPMDB_GPG_PUBKEY_REMOVAL` looks like you're talking about a cmake flag, i.e. for `#ifdef`. I want a runtime flag for repo_add_rpmdb. Hmm, so it should probably be `RPMDB_KEEP_GPG_PUBKEY`.

j-mracek:
Originally I planed it as cmake flag, but probably bag idea.

j-mracek:
Thanks

j-mracek:
I have discovered that it will requires more deeper changes in libdnf, therefore I am closing it.

PR #270, opened 2018-06-12
Add zchunk support

jdieter:
This patch adds [zchunk](https://github.com/zchunk/zchunk) support to libsolv.


CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 PROJECT (libsolv)
 
 CMAKE_MINIMUM_REQUIRED (VERSION 2.4)
+IF (COMMAND CMAKE_POLICY)
+CMAKE_POLICY(SET CMP0003 NEW)
```
ignatenkobrain: why?

jdieter: If I don't, I get the following:
```
CMake Warning (dev) at ext/CMakeLists.txt:137 (ADD_LIBRARY):
  Policy CMP0003 should be set before this line.  Add code such as

    if(COMMAND cmake_policy)
      cmake_policy(SET CMP0003 NEW)
    endif(COMMAND cmake_policy)

  as early as possible but after the most recent call to
  cmake_minimum_required or cmake_policy(VERSION).  This warning appears
  because target "libsolvext" links to some libraries for which the linker
  must search:

    zck

  and other libraries with known full path:

    /home/jonathan/Documents/programming/libsolv/build/src/libsolv.so.0

  CMake is adding directories in the second list to the linker search path in
  case they are needed to find libraries from the first list (for backwards
  compatibility with CMake 2.4).  Set policy CMP0003 to OLD or NEW to enable
  or disable this behavior explicitly.  Run "cmake --help-policy CMP0003" for
  more information.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

ignatenkobrain: I think this might indicate wrong pkg-config file for zck, however I can't test it since zck is not packaged for Fedora.

jdieter: There is a copr at https://copr.fedorainfracloud.org/coprs/jdieter/zchunk/, but, you're right, we really need to get this into Fedora.

Neal, are you up for a package review?

ignatenkobrain: Just assign me for review.

jdieter: Great, thank you!

Conan-Kudo: I'm ready to review as well.

jdieter: Ok, Igor, I've assigned it to you, and, Neal, I've cc'd you into the bug.  Thanks so much, both of you.

ignatenkobrain: I don't see any bugs assigned to i.gnatenko.brain@gmail.com

jdieter: Sorry, I assigned it to ignatenko@redhat.com since that's what it looked like your bugzilla account was.  I'll re-assign it.

mlschroe:
(But doesn't all this belong in the transport layer, aka librepo? Why do you need that in libsolv at all?)

jdieter:
Zchunk is a new compression format that libsolv needs to be able to decompress, just as it decompresses xz and gz files.

mlschroe:
I'll do a little zchunk review today. Stay tuned.

jdieter:
Thanks so much!

mlschroe:
Ok, see my mail to rpm-ecosystem@lists.rpm.org
Thanks!

Conan-Kudo:
@mlschroe Why'd you implement it by hand in 75d03059cdaf7770d2217fafaba6af732cd860ba instead of pulling this PR in and using the library?

The library is available in openSUSE and can be easily backported to SLE 12/Leap 42 if needed.

jdieter:
@mlschroe If there's anything missing in the library, I would happily add it, and, if you're worried about control, I'll happily give you commit access.  I'm just concerned that, as new features are added to the library, they'll also have to be manually added to libsolv before we can take advantage of them.

mlschroe:
Because it was fun to do and got me a good understanding for the format. And I somewhat don't like dragging in a dependency to some other library/libraries for just so little lines of code. Anyway, I can make this configurable if you like.

ignatenkobrain:
@jdieter can you rebase this please and add configurable switch to use system zchunk? ;)

jdieter:
Ok, I've rebased this and rewritten it to use a new switch, USE_SYSTEM_ZCHUNK.  I've tested with the switch both enabled and disabled, and it seems to work perfectly.

CMakeLists.txt
```diff
@@ -156,6 +159,12 @@ IF (ENABLE_ARCHREPO OR ENABLE_DEBIAN)
 SET (ENABLE_LZMA_COMPRESSION ON)
 ENDIF (ENABLE_ARCHREPO OR ENABLE_DEBIAN)
 
+IF (USE_SYSTEM_ZCHUNK)
+SET (ENABLE_ZCHUNK_COMPRESSION ON)
```
Conan-Kudo: This check should be inverted. It should be `USE_SYSTEM_ZCHUNK` underneath `ENABLE_ZCHUNK_COMPRESSION`.

jdieter: I think it's right the way it is.  It's saying that if USE_SYSTEM_ZCHUNK is enabled, we should automatically enable ENABLE_ZCHUNK_COMPRESSION.  However, if you're using the embedded zchunk decompressor, ENABLE_ZCHUNK_COMPRESSION should be enabled, and USE_SYSTEM_ZCHUNK shouldn't be.

mlschroe:
No, it's fine as it is. I'll merge it tomorrow.

mlschroe:
(for a very loose definition of tomorrow...)

jdieter:
That works!  Thanks so much!

mlschroe:
I did some changes to the code:

commit 9b413e2c3790ca11eb89debffe881d43856e0bc1 (HEAD -> master, origin/master)
Author: Michael Schroeder <mls@suse.de>
Date:   Mon Oct 8 11:55:13 2018 +0200

    Tweak system zchunk code
    
    - rename USE_SYSTEM_ZCHUNK to WITH_SYSTEM_ZCHUNK
    - fix file descriptor leak
    - fix memory leaks

Could you please check that is still works? Thanks!

jdieter:
I've just done a pull request, #282, that fixes some very small bugs in 9b413e2

mlschroe:
Thank you!

PR #273, opened 2018-06-21
Bump debian control.tar size limit

dbnicholson:
Increase the arbitrary control.tar.gz limit from 1 MB to 16 MB. We came
across an openjdk-7-jdk package that had a 2.4MB control.tar.gz and
libsolv failed to parse it. dpkg doesn't actually apply a limit to these
files.


mlschroe:
Yes, the limit is somewhat arbitrary. It's just there to guard against bad packages. (I'm surprised that dpkg doesn't have limit, rpm does have limits for header sizes.)

PR #283, opened 2018-10-08
add .clang-format

ignatenkobrain:
So that it is possible to automatically format code.

---

@mlschroe I was not sure about indentation, but it seems that libsolv is not formatted correctly everywhere.

At some point I would like to make sure that clang-format applies to libsolv source code without any changes, but this is very close.


wolfv:
I also noticed quite a few tab / space mixes.

Issue #258, opened 2018-03-02
RFE: expose "considered" in bindings


ignatenkobrain:
This probably will need some work on exposing `map*` too.

ignatenkobrain:
FTR, @dparalen is working on this.

dparalen:
yeah, early stages tho O:-)

dparalen:
@ignatenkobrain hey man! :)
So finally got some time to start looking into this. IIRC you recommended me exposing the `struct Map` in some generic way. So here's a brief (stand-alone) SWIG bindings file exposing that:

```swig
// bitmap.i
// SWIG interface to the libsolv bitmap object

%module bitmap
%{
  #include "bitmap.h"
  #include "pooltypes.h"
%}

%include "bitmap.h"

typedef int Id;

%typemap(freearg) Map {
  map_free($1);
}

%extend _Map {
  void set_id(Id id) {
    // If encountered an ID bigger than size: resize the mapping first for the
    // ID to fit. The initial `$self->size == 0` is cool thanks to the handling
    // in bitmap:map_grow ~> util:solv_realloc. map_grow checks whether a
    // resize is actually needed.
    map_grow($self, (int)id);
    map_set($self, (int)id);
  }
}

%extend _Map {
  Id test_id(Id id){
    map_grow($self, (int)id);
    return (Id) map_tst($self, (int)id);
  }
}

%extend _Map {
  void clear_id(Id id){
    map_grow($self, (int)id);
    map_clr($self, (int)id);
  }
}
```
This can be compiled and "tested" in python:
* `cd libsolv && mkdir tmp && cd tmp`
* `ln -s ../src/bitmap.c ../src/bitmap.h ../src/util.c ../src/util.h ../src/pooltypes.h ./`
* supposing the above binding is saved locally: `swig -python bitmap.i`
* I've been testing on Fedora28 with Python2 (Pulp) so: `gcc -fpic -DHAVE_QSORT_R util.c bitmap.c bitmap_wrap.c -I/usr/include/python2.7`
* `gcc -shared  bitmap.o bitmap_wrap.o util.o -o _bitmap.so`

Having done that, one can play with the binding:
```python
import bitmap
m = bitmap.Map()
m.set_id(1024)
m.size
```
So I'd like to get some early feedback before I dig into making the `map->size` and `map->map` attributes read only etc. What I'm wondering is whether to use the `Id` type since it can be negative inside the solver... Also whether a `Map` makes sense outside of a `Pool`...

[...]

PR #288, opened 2018-11-09
Expose pool.set_/get_considered in Python bindings

dparalen:
This patch exposes pool.considered bitmap thru a pair of get/set methods.

Fixes: #258

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -1131,6 +1131,11 @@ typedef struct {
   int const count;
 } TransactionClass;
 
+typedef struct {
+  unsigned char *map;
+  int size;
+} Map;
+
 %extend SolvFp {
   ~SolvFp() {
     if ($self->fp)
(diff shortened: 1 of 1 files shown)


mlschroe:
Hmm, I think this is way too complicated for what it does. But I do like the idea of using a "list of set bits". So let's do it like this: We'll add:

```
pool.set_considered_list(list)
list = pool.get_considered_list()
```

That's basically the interface of your helpers. But we do not need a typemap for this, we can simply use the Queue typemap to implement those two functions. (I.e. it will also work for perl/ruby/tcl)

dparalen:
@mlschroe with the update, the interface looks like this in Python:

```
In [1]: import solv

In [2]: p = solv.Pool()

In [3]: s = p.add_repo('r').add_solvable()

In [4]: p.get_considered()
Out[4]: []

In [5]: p.set_considered([0, 0, 1])

In [6]: p.get_considered()[s.id] & 1
Out[6]: 0

In [7]: p.get_considered()[s.id] != 0
Out[7]: True

In [8]: p.get_considered()[s.id] & 7
Out[8]: 4

In [9]: p.set_considered([])

In [10]: p.get_considered()
Out[10]: []
```

ignatenkobrain:
Any reason not to use True/False?

dparalen:
might have to implement a language-specific conversion to the target language array then because `True/False` are objects in e.g Python.

How about exposing some `def set_considered_pythonic()` method that would take care about this and could work with e.g sets instead?

ignatenkobrain:
```c
%typemap(out) bool {
   
    if($1)
        $result = (PyObject *)Py_True;
    else
        $result = (PyObject *)Py_False;
}
```

Won't something like this work?

dparalen:
Actually, never mind me O:-) updated with the additional conversion:

```
In [1]: import solv                                                                                                                                                                                                
                                                                                                                                                                                                                   
In [2]: p = solv.Pool()                                                                                                                                                                                            
                                                                                                                                                                                                                   
In [3]: s = p.add_repo('r').add_solvable()                                                                                                                                                                         
                                                                                                                                                                                                                   
In [4]: p.set_considered([False, 0, True])                                                                                                                                                                         
                                                                                                                                                                                                                   
In [5]: p.get_considered()                                                                                                                                                                                         
Out[5]: [False, False, True, False, False, False, False, False]                                                                                                                                                    
                                                                                                                                                                                                                   
In [6]: p.set_considered([])                                                                                                                                                                                       
                                                                                                                                                                                                                   
In [7]: p.get_considered()                                                                                                                                                                                         
Out[7]: []                                                                                                                                                                                                         
```

mlschroe:
Before we do more comment ping/pong I committed something that can act as a base for further improvements. I didn't support the "reset" operation because I don't like to tie it to the empty list. Maybe we need another method for it.

dparalen:
> Hmm, should have looked at the pull request first. So just rename them ;)

lol :D I absolutely didn't notice the name of the methods you were suggesting before, sorry about that

dparalen:
> Before we do more comment ping/pong I committed something that can act as a base for further improvements. I didn't support the "reset" operation because I don't like to tie it to the empty list. Maybe we need another method for it.

Cool, thanks for the patch! Yeah when it comes to the reset, since empty list has the same semantics as everything considered (or doesn't it?) I thought it was OK to use it as the reset "flag" but I'm cool with any other way.

dralley:
@mlschroe I could be doing something wrong, but I cannot get these methods to work.

If I call ```get_considered_list()``` without first using ```set_considered_list()```, it works.

Screenshot from 2019-06-15 18-59-17

If I call ```set_considered_list([True] * 45)```, (there are 45 solvables in the list), then calling ```get_considered_list()``` afterwards is broken.

Screenshot from 2019-06-15 18-02-10

Am I misusing it?  Or is there a memory bug somewhere?  Depsolving was broken after having set the considered list as described above.

mlschroe:
Commit e33bb18fc9205d79228503e13f7a2b6502dd7675 also adds pool.set_disabled_list() and pool.get_disabled_list().

mlschroe:
Closing. Please reopen if something is still missing.
(I'll also document the calls in the bindings manpage)

PR #296, opened 2019-02-06
Add support for modular updateinfo.xml data

j-mracek:
Module advisory data can specify which advisory is applicable for which
modular stream. Module information can be absent or there could be
multiple entries.

If there is a module entry in advisory and none of modules in advisory
is enabled on the system, advisory is not relevant for the system and
should be ignored.

### DIFF
--- a/ext/repo_updateinfoxml.c
+++ b/ext/repo_updateinfoxml.c
@@ -37,6 +37,7 @@
  *     <pkglist>
  *       <collection short="F8">
  *         <name>Fedora 8</name>
+ *         <module name="pki-deps" stream="10.6" version="20181019123559" context="9edba152" arch="x86_64"/>
  *         <package arch="ppc64" name="imlib-debuginfo" release="6.fc8" src="http://download.fedoraproject.org/pub/fedora/linux/updates/8/ppc64/imlib-debuginfo-1.9.15-6.fc8.ppc64.rpm" version="1.9.15">
  *           <filename>imlib-debuginfo-1.9.15-6.fc8.ppc64.rpm</filename>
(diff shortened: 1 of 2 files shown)


j-mracek:
@ignatenkobrain I am not sure what exactly you suggest to change. Do you want to rename:
UPDATE_MODULE to UPDATE_COLLECTION_MODULE
UPDATE_MODULE_NAME to UPDATE_COLLECTION_MODULE_NAME
UPDATE_MODULE_STREAM to UPDATE_COLLECTION_MODULE_STREAM
UPDATE_MODULE_VERSION to UPDATE_COLLECTION_MODULE_VERSION
UPDATE_MODULE_CONTEXT to UPDATE_COLLECTION_MODULE_CONTEXT
UPDATE_MODULE_ARCH to UPDATE_COLLECTION_MODULE_ARCH

Also I am not sure if note ``moreover, it should create a conflict entry with old modular package rather then just storing data`` is applicable. Please see implementation in libdnf how module advisory data are used.

I also would like to mentioned that implementation of parsing of the new modular element in update collection is important for DNF team.

ignatenkobrain:
No, what I meant is:

create a `patch:…` solvable which conflicts with old versions of modules as it is done for RPMs now.

mlschroe:
I think the pull request is fine as it is, except that you must append the new knownid entries to the end of the list so that the ABI is unchanged. The next time we break the ABI we can move them to a better place.

mlschroe:
(UPDATE_COLLECTION is misnamed, it should have been UPDATE_PACKAGE)

j-mracek:
@mlschroe Part in knownid was moved. Thanks for review.

PR #301, opened 2019-02-23
Not considered excluded packages as a best candidate

j-mracek:
In some work-flows we can end up in selection of a best candidate with
a package that is not in considered map, therefore it cannot be
installed. Removal of those packages as a first prune process solves
the issue.

https://bugzilla.redhat.com/show_bug.cgi?id=1677583
https://bugzilla.redhat.com/show_bug.cgi?id=1678911
https://bugzilla.redhat.com/show_bug.cgi?id=1673851

### DIFF
--- a/src/policy.c
+++ b/src/policy.c
@@ -831,6 +831,33 @@ move_installed_to_front(Pool *pool, Queue *plist)
     }
 }
 
+
+/*
+ * prune_to_considered
+ *
+ * Keep only considered (pool->considered) solvables in plist. If pool->considered is NULL, it keeps
+ * all solvable in plist. If non of solvables in plist is in considered map, it returns empty plist.
+ */
+static void
+prune_to_considered(Pool *pool, Queue *plist)
+{
+  if (plist->count == 0)		/* no need to prune if plist is empty */
+    return;
+  if (!pool->considered)        /* no need to prune if no considered map */
+    return;
+  int i, j;
+  Id id;
+  for (i = j = 0; i < plist->count; i++)
+    {
+      id = plist->elements[i];
+      if (MAPTST(pool->considered, id))
+        {
+          plist->elements[j++] = id;
+        }
+    }
+  plist->count = j;
+}
+
 /*
  * prune_to_best_version
  *
@@ -1299,6 +1326,7 @@ policy_filter_unwanted(Solver *solv, Queue *plist, int mode)
 	  return;
 	}
     }
+  prune_to_considered(pool, plist);
   if (plist->count > 1)
     {
       if (mode != POLICY_MODE_SUGGEST)


src/policy.c
```diff
@@ -831,6 +831,32 @@ move_installed_to_front(Pool *pool, Queue *plist)
...
 }
 
+
+/*
+ * prune_to_considred
+ *
+ * Keep only considered
+ */
+void
+prune_to_considred(Pool *pool, Queue *plist)
```
Conan-Kudo: Don't you mean `considered`? Let's not put typos into the API. 😉

src/policy.c
```diff
@@ -831,6 +831,32 @@ move_installed_to_front(Pool *pool, Queue *plist)
     }
 }
 
+
+/*
+ * prune_to_considred
+ *
+ * Keep only considered
```
Conan-Kudo: Uhh? This is a very useless description. What does this _do_?

j-mracek:
@Conan-Kudo Thanks. Is it better now?

Conan-Kudo:
@j-mracek Code looks good to me.

src/policy.c
```diff
@@ -831,6 +831,33 @@ move_installed_to_front(Pool *pool, Queue *plist)
     }
 }
 
+
+/*
+ * prune_to_considred
```
Conan-Kudo: Oops, typo here, too. Please fix. :)

j-mracek: @Conan-Kudo Thanks again, typo was fixed. You know I was so exited that it solves our critical issue that I overlooked problems in description and def name.

src/policy.c
```diff
@@ -1299,6 +1326,7 @@ policy_filter_unwanted(Solver *solv, Queue *plist, int mode)
 	  return;
 	}
     }
+  prune_to_considered(pool, plist);
```
ignatenkobrain: this call should be guarded by `if plist->count > 1`

src/policy.c
```diff
@@ -831,6 +831,33 @@ move_installed_to_front(Pool *pool, Queue *plist)
     }
 }
 
+
+/*
+ * prune_to_considered
+ *
+ * Keep only considered (pool->considered) solvable in plist. If pool->considered is NULL it keeps
```
ignatenkobrain: typo: `solvables`

ignatenkobrain: `s/it keeps/keep/`

src/policy.c
```diff
@@ -831,6 +831,33 @@ move_installed_to_front(Pool *pool, Queue *plist)
...
+ * all solvable in plist.
```
ignatenkobrain: `solvables`

src/policy.c
```diff
@@ -831,6 +831,33 @@ move_installed_to_front(Pool *pool, Queue *plist)
...
+ */
+static void
+prune_to_considered(Pool *pool, Queue *plist)
+{
+  if (plist->count < 2)		/* no need to prune for a single entry */
```
ignatenkobrain: I don't think this belongs to this function.

ignatenkobrain:
code style differs from the rest of policy.c, please fix it (I'm talking about indentation)

src/policy.c
Conan-Kudo: This is actually grammatically correct.

src/policy.c
Conan-Kudo: It probably makes sense, since that prevents it from filtering to zero on accident.

src/policy.c
j-mracek: The test is part of the function

j-mracek:
> code style differs from the rest of policy.c, please fix it (I'm talking about indentation)

Please can you be more specific. I tried to use exactly the same style like in other functions.

src/policy.c
j-mracek: Thanks for pointing here, because for correct functionality the condition should be ``if (plist->count == 0)``. It is really required to remove all excluded solvables for update job even if it ends up in empty plist. The install job will report (will only not mention the best rule)
```
Problem: conflicting requests
  - package TestA-1.0.0-1.noarch is excluded
  - package TestA-1.0.0-2.noarch is excluded
```

mlschroe:
I don't think this makes sense. How do non-considered packages end up on the list anyway?

j-mracek:
> I don't think this makes sense. How do non-considered packages end up on the list anyway?

We changed workflows in ``libdnf/dnf_sack.cpp`` in def ``dnf_sack_recompute_considered(DnfSack *sack)`` where we recompute provides (``pool_createwhatprovides(priv->pool)``) before removing excluded packages from ``pool->considered`` Map. We changed it because we want to preserve the meaning of ``ID``s generated by ``pool_queuetowhatprovides(dnf_sack_get_pool(pImpl->sack), pkgs.getQueue())`` because we call ``pool_createwhatprovides(priv->pool)`` only once. 

Additionally we were able to get information about excluded packages from solver like ``package TestA-1.0.0-2.noarch is excluded`` even when it is as a dependency (provides of excluded packages are visible for solver).

The change also provides some problems like with definition of best candidate in case when we use solver_solvable_all or solver_solvable_provides type (more) in job.

The patch solves our issue and as you mentioned in other workflows ``plist`` has not excluded packages therefore it should be not a harmful to zipper or so on.

mlschroe:
Sorry, but you mustn't do that. The whatprovides list needs to be in sync with the considered packages.

mlschroe:
Things will break if all considered providers are in conflict, as the solver will not try to find the best package in that case. And there are probably other cases where your method will not work.

(I think I also mentioned that some weeks ago...)

mlschroe:
(Note that I did say some years ago that I feel using the considered packages mechanism for exclude packages is a mistake. SOLVER_LOCK was meant to be used for this.)

mlschroe:
(regarding the "Things will break..." comment: that was not true, because the solver will create a "conflict" for the packages that are not considered. But I really don't know if there are other areas that will break. The whatprovides hash not matching the considered packages is pretty much an unsupported configuration.)

Conan-Kudo:
@mlschroe The problem seems to be that libsolv isn't creating absolute conflicts when using `SOLVER_LOCK` to lock packages. When `forcebest` flag is set, it seems to override locked packages and select them anyway.

To me, this seems like broken behavior, because I expect that locking packages makes them permanently excluded from the pool for solver calculations (that is, they don't exist).

Conan-Kudo:
@mlschroe Can you please look at this and #302?

j-mracek:
@mlschroe Thanks a lot

mlschroe:
Hey, I added a `POOL_FLAG_WHATPROVIDESWITHDISABLED` pool flag! You can use this to tell libsolv that the whatprovides hash should include disabled packages. Thus not calling pool_createwhatprovides() after changing the considered map is no longer unsupported.

I didn't do much testing, please give it a try.

j-mracek:
@mlschroe Thanks, I will test it.

j-mracek:
@mlschroe It works perfectly. Thanks a lot.

PR #303, opened 2019-03-12
Add zlib include directories

wolfv:


ignatenkobrain:
Seems fair, don't we need to do same for others?

wolfv:
yep, i think that would be better. It was the only issue coming up for building `libsolv` for conda which installs all dependencies into a different root.

wolfv:
btw how portable is libsolv? Just *if* the move to use libsolv in conda grows momentum, it would be necessary to compile on Windows as well.

So far it seems to fail: https://github.com/conda-forge/staged-recipes/pull/7969

Conan-Kudo:
@wolfv I think the only thing is we need some alternative to `fopencookie(3)` in Windows.

wolfv:
@ignatenkobrain I added some more. Since zstd uses `find_library` (not `find_package`) and I don't have it installed, I didn't add a include_directories directive there.
I think it should read something like 

```
find_path(
    ZSTD_INCLUDE_DIR
    NAMES "zstd.h"
    HINTS
        "/usr/local/facebook/include"
)
```

(taken from https://github.com/facebookarchive/beringei/blob/master/beringei/cmake/FindZStd.cmake)

Conan-Kudo:
@wolfv It could be converted to use `pkg_check_modules` for this, right? it ships a libzstd.pc file...

wolfv:
let's rather do the find_path as it's done in libsolv's findLZMA

https://github.com/openSUSE/libsolv/blob/master/cmake/modules/FindLZMA.cmake#L13

Conan-Kudo:
@wolfv Sure, feel free to introduce a `FindZSTD.cmake` file. :)

PR #306, opened 2019-03-19
Adding MSVC support

wolfv:
I've added a win32_compat folder containing necessary headers to allow building libsolv on MSVC successfully. 
I also added CI with appveyor: https://ci.appveyor.com/project/wolfv/libsolv

Most of the polyfills come from `musl` with slight changes to make them compile with MSVC.

`pread.c` and `wintime.h` are ported from postgresql sources.

I'll probably need to clean this up a little bit, but I would be very happy about a review. I am not a C expert yet and do not really know many of these low level interfaces.

The one remaining bigger problem (in repopage.c) is 

```
    fcntl(store->pagefd, F_SETFD, FD_CLOEXEC);
```

which I've just disabled completely for now. I don't think that's correct, but I also don't know what this function does. Can someone help?


SylvainCorlay:
Awesome!

CMakeLists.txt
```diff
@@ -2,8 +2,14 @@ PROJECT (libsolv)
...
 
+IF (MSVC)
+  SET (DISABLE_SHARED_DEFAULT ON)
+ELSE ()
+  SET (DISABLE_SHARED_DEFAULT OFF)
+ENDIF ()
+
 OPTION (ENABLE_STATIC "Build a static version of the libraries?" OFF)
-OPTION (DISABLE_SHARED "Do not build a shared version of the libraries?" OFF)
+OPTION (DISABLE_SHARED "Do not build a shared version of the libraries?" DISABLE_SHARED_DEFAULT)
```
Conan-Kudo: Why are we making this change? Doesn't it make more sense just to set the option accordingly for an individual build?

CMakeLists.txt
```diff
@@ -420,13 +431,16 @@ SET (SYSTEM_LIBRARIES "-Wl,--as-needed" ${SYSTEM_LIBRARIES})
...
+
+IF (NOT MSVC)
+  ADD_SUBDIRECTORY (ext)
+  ADD_SUBDIRECTORY (tools)
+  IF (ENABLE_PERL OR ENABLE_PYTHON OR ENABLE_RUBY OR ENABLE_TCL)
+      ADD_SUBDIRECTORY (bindings)
+  ENDIF (ENABLE_PERL OR ENABLE_PYTHON OR ENABLE_RUBY OR ENABLE_TCL)
+  ADD_SUBDIRECTORY (examples)
+  ADD_SUBDIRECTORY (doc)
+ENDIF ()
```
Conan-Kudo: Why are we saying none of these can be built with MSVC? Is that actually true?

src/CMakeLists.txt
```diff
@@ -29,7 +29,22 @@ SET (libsolv_HEADERS
...
+        win32_compat/fnmatch.c
+        win32_compat/getopt.c
+        win32_compat/pread.c
+        win32_compat/regcomp.c
+        win32_compat/regexec.c
+        win32_compat/strcasecmp.c
+        win32_compat/strcasestr.c
+        win32_compat/tre-mem.c
+    )
+ENDIF ()
```
Conan-Kudo: This extra stuff isn't for just MSVC, this is for Win32 as a whole. Do the right thing here, please.

src/util.c
```diff
@@ -11,7 +11,11 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <string.h>
+#ifndef _WIN32
 #include <sys/time.h>
+#else
+#include "wintime.h"
```
Conan-Kudo: Can't we invert this logic for readability?

Conan-Kudo:
@wolfv Please add a license file for the musl-derived code in the `win32_compat` folder.

Conan-Kudo:
@wolfv Also, it might be more appropriate to have `win32_compat` as a top level directory rather than being under `src`, since it could also be used for `libsolvext` too...

CMakeLists.txt
wolfv: Building the `ext` seemed to be quite hard due to fncntl ... I don't have the time + knowledge to look into it right now, unfortunately. I think doing a first, small PR to get something to build successfully on Windows is a better approach.

src/CMakeLists.txt
wolfv: Sure, I can change that.

Conan-Kudo:
@wolfv Could you rebase onto current master please?

win32_compat/README
```diff
@@ -0,0 +1,28 @@
...
+  - fnmatch
+  - getopt
+  - regerror
+  - regex
+  - strcasecmp/strncasecmp
+  - strcasestr
+
+- stackoverflow (CC-BY-SA)
+  - unistd (https://stackoverflow.com/a/826027/1347553)
+  - pread (https://stackoverflow.com/a/42495892)
```
Conan-Kudo: I'm unsure if CC-BY-SA is compatible with all the downstream users' licensing...

wolfv:
Alright!

win32_compat/README
```diff
@@ -0,0 +1,28 @@
...
+
+- musl / TRE (BSD3 License)
```
Conan-Kudo: Please use SPDX notation for this particular license, as `BSD3` is ambiguous...

CMakeLists.txt
```diff
@@ -362,61 +371,86 @@ INCLUDE(CPack)
...
 ADD_SUBDIRECTORY (src)
-ADD_SUBDIRECTORY (ext)
-ADD_SUBDIRECTORY (tools)
-IF (ENABLE_PERL OR ENABLE_PYTHON OR ENABLE_RUBY OR ENABLE_TCL)
+
+IF (NOT WIN32)
+
+  OPTION (ENABLE_TOOLS "Build libsolv tools (linux only)" ON)
+  OPTION (ENABLE_EXT "Build libsolv extensions (linux only)" ON)
+  OPTION (ENABLE_EXAMPLES "Build libsolv examples (linux only)" ON)
```
Conan-Kudo: Why are these marked as Linux only? That's definitely not true. It likely wouldn't even require much to get it to work on Windows, but it works on multiple platforms today.

ext/CMakeLists.txt
```diff
@@ -13,146 +13,162 @@ ENDIF (ENABLE_RPMDB OR ENABLE_RPMPKG)
...
+    TARGET_LINK_LIBRARIES(libsolvext libsolv ${SYSTEM_LIBRARIES})
 ENDIF (DISABLE_SHARED)
 
-SET_TARGET_PROPERTIES(libsolvext PROPERTIES OUTPUT_NAME "solvext")
-SET_TARGET_PROPERTIES(libsolvext PROPERTIES SOVERSION ${LIBSOLVEXT_SOVERSION})
-SET_TARGET_PROPERTIES(libsolvext PROPERTIES INSTALL_NAME_DIR ${CMAKE_INSTALL_LIBDIR})
+SET_TARGET_PROPERTIES(libsolvext
+    PROPERTIES
+      OUTPUT_NAME "solvext"
+      MACOSX_RPATH ON
```
Conan-Kudo: Why do we want RPATH on? Usually we'd prefer to have it off...

src/CMakeLists.txt
```diff
@@ -33,28 +32,60 @@ IF (ENABLE_CONDA)
...
+    ADD_LIBRARY (libsolv SHARED ${libsolv_SRCS})
+ENDIF ()
 
-SET_TARGET_PROPERTIES(libsolv PROPERTIES OUTPUT_NAME "solv")
-SET_TARGET_PROPERTIES(libsolv PROPERTIES SOVERSION ${LIBSOLV_SOVERSION})
-SET_TARGET_PROPERTIES(libsolv PROPERTIES INSTALL_NAME_DIR ${CMAKE_INSTALL_LIBDIR})
+SET_TARGET_PROPERTIES(libsolv
+    PROPERTIES
+      OUTPUT_NAME "solv"
+      MACOSX_RPATH ON
```
Conan-Kudo: Why do we want RPATH on? Usually we'd prefer to have it off...

wolfv:
Great! Tests seem to run perfectly fine on OS X as well. 

Here is the appveyor build: https://ci.appveyor.com/project/wolfv/libsolv if some core dev enables them in here, this will be tested upon PR as well. Currently, it only tests compilation, since making `ext` build is quite cumbersome.

CMakeLists.txt
wolfv: You're right, they should be `not win`. 

I think it requires quite a bit of effort to get them to build, due to use of (mainly) fcntl for which I haven't found the appropriate substitute in Windows. (I am also no expert, so please feel free to correct me if I am wrong!)

win32_compat/README
wolfv: makes sense.

win32_compat/README
wolfv: Hmm. We could remove unistd completely (before I was commenting it out everywhere and I think libsolv will build without it.
For pread we can also take the one from postgres sources and modify it.

ext/CMakeLists.txt
wolfv: Let's try to remove it. I needed it for the conda-forge build to be able to properly link against it but this might be gone now that we're using the GNU Install Dirs.
I'm no expert when it comes to OS X and rpath handling. In the worst case I can probalby set a default value on the command line for the OS X build or patch the cmake.

CMakeLists.txt
Conan-Kudo: There's an implementation of `fcntl` in newlib, would that work?

Also, `fopencookie` exists in musl now, so that could help with libsolvext...

CMakeLists.txt
wolfv: Yep, you're right.

CMakeLists.txt
wolfv: Hmm, I am taking this back. I'll try to get ext to build on Windows as well. seems like just a replacement for `sys/stat.h` is needed, and that seems to be `direct.h` on Windows.

wolfv:
I've tried to build `ext` on Windows, but it's consuming too much of my time right now. It's not entirely trivial to get the fopencookie/funopen replacement working. I found the fmem library, which might help, but doesn't have drop-in replacement (https://github.com/Snaipe/fmem)

A lot of libraries implement fopencookie etc. but rely on system calls for the functionality. 
Again, I am no expert in either Windows C and WIN 32 api or linux, so it's really hard for me to judge how to get this functionality right.

If possible, I would love to get this PR in first (with just core-library functionality), and then at a later stage make sure that `ext` and the tests are building fine as well. 

This PR also adds the necessary CI infrastructure for continuous testing on OS X and Windows, so I think there is a lot of value here. 

@mlschroe what's your take on the `win32_compat` folder? And the `cc-by-sa`?

mlschroe:
I don't think cc-by-sa is compatible with the bsd license libsolv uses. I also don't think it makes sense to plug in musl's stdio implementation, as it's most likely not compatible to MSVC's stdio implementation. So no fopencookie I guess.

wolfv:
are you ok with the `win32_compat` folder in general? And are you ok to merge this PR once I've replaced the CC-BY-SA files with either my implementations or another BSD3 implementations, or is this PR hopeless?

This would make me able to remove the patches I currently have in the libsolv package for conda.

mlschroe:
Regarding the CLOEXEC flag: on windows you probably want to clear the HANDLE_FLAG_INHERIT flag instead: `SetHandleInformation(handle, HANDLE_FLAG_INHERIT, 0)`

mlschroe:
Oh, I' wouldn't say it is hopeless ;)

You probably also don't need gettimeofday() implementation, just add the code to solv_timems.
(And wouldn't simply calling GetSystemTimeAsFileTime() and then converting to millisecs be enough?)

mlschroe:
There's also GetTickCount() which seems to do exactly what we need.

mlschroe:
See also QueryPerformanceFrequency() and QueryPerformanceCounter() if we need a higher resolution (the tick count is only updated every 10msec)

mlschroe:
Commit a7bba7275e762788e39b3c095fd9f3ec16ab1169 adds the option to build without stio cookie opens. You still need fmemopen() for testcase support, though.

wolfv:
great :pray: I'll take another look tonight.

mlschroe:
I added a `solv_setcloexec()` function in util.c so that you just have to modify one place in the code.

Conan-Kudo:
@wolfv Could this help for making an fopencookie implementation compatible with the standard Windows stdio? https://gitlab.com/Conan_Kudo/portable-fopencookie

wolfv:
Thanks everybody! I have been successful in running the tests using an implementation of fmemopen on Windows. I'll check out the portable-fopencookie, but I'll probably leave it at fmemopen for now, if that's ok for everyone. I'm working on putting everythign together, will push shortly.

wolfv:
This is now somewhat successfully running the tests:
https://ci.appveyor.com/project/wolfv/libsolv

CMake seems to complain about escape sequences due to paths' like `C:\path` on windows (note the backslash), and there are some `6: testcase_read: cannot parse command '=Con:'` which I can't explain (also did not yet further debug).

ignatenkobrain:
```
  testcasedir = solv_strdup(testcase);                                                                        
  if ((s = strrchr(testcasedir, '/')) != 0)                                                                   
    s[1] = 0;                                                                                                 
  else                                                                                                        
    *testcasedir = 0;                                                                                         
```

It might be related to this.

ext/testcase.c
```diff
@@ -2573,7 +2580,7 @@ testcase_read(Pool *pool, FILE *fp, const char *testcase, Queue *job, char **res
       return 0;
     }
   testcasedir = solv_strdup(testcase);
-  if ((s = strrchr(testcasedir, '/')) != 0)
+  if ((s = strrchr(testcasedir, PATH_SEP)) != 0)
```
wolfv: @ignatenkobrain I did change that here

mlschroe:
I hope I fixed that ctest warning issue with commit 3726508400bee9980c05633f8e0339e41c4921ef

mlschroe:
For fmemopen something like this should do (untested):

```
FILE *
fmemopen(void *buf, size_t size, const char *mode)
{
  char tempath[MAX_PATH + 1];
  char tempnam[MAX_PATH + 1];
  DWORD l;
  HANDLE fh;
  FILE *fp;

  if (strcmp(mode, "r") != 0 && strcmp(mode, "r+") != 0)
    return 0;
  l = GetTempPath(MAX_PATH, temppath);
  if (!l || l >= MAX_PATH)
    return 0;
  if (!GetTempFileName(temppath, "solvtmp", 0, tempnam))
    return 0;
  fh = CreateFile(tempnam, DELETE | GENERIC_READ | GENERIC_WRITE, 0,
                  NULL, CREATE_ALWAYS,
                  FILE_ATTRIBUTE_NORMAL | FILE_FLAG_DELETE_ON_CLOSE,
                  NULL);
  if (fh == INVALID_HANDLE_VALUE)
    return 0;
  fp = win32_fdopen(win32_open_osfhandle((intptr_t)fh, 0), "w+b");
  if (!fp)
    return 0;
  if (buf && size && fwrite(buf, size, 1, fp) != 1)
    {
      fclose(fp);
      return 0;
    }
  rewind(fp);
  return fp;
}

```

Maybe there is some way to implement it without using a tmpfile, but I haven't found it yet.

wolfv:
Thanks, making another pass now! (btw. I already added fmemopen replacement here: https://github.com/wolfv/libsolv/blob/msvc_support/win32_compat/fmemopen.c but I'll try replacing it with your version).

mlschroe:
Looks pretty similar. Oh well.

mlschroe:
Btw, could you please remove all the CMakeLists.txt indenting changes from the commits? While it's a good idea to fix the indents it should not be part of the win32 port.

wolfv:
If you insist, I can do that :) Running clang-format would be a great help, too, btw. Currently there are a lot of mixed tabs + spaces which is making indentation look weird.

wolfv:
btw. is it OK to remove unistd almost everywhere? It seems to compile fine without it on Fedora 29 but not sure about older systems?

wolfv:
Looks like I successfully got rid of all non-musl dependencies. That will make the licensing issue simpler, hopefully. Next, I'll revert the non-essential changes to the CMake files, and then I hope this is mostly good to go!

mlschroe:
Hmm, I don't think we can drop unistd.h for all distros. It's supposed to contain lots of POSIX definitions. Just put an empty unistd.h in the win32 dir.

Btw, could you please rename `win32_compat` to simply `win32`? Thanks!

wolfv:
Ok, I've now reverted the cmake formatting to be as close to the previous state as possible, and renamed the folder to `win32`.

Remaining items:

- Fix the fcntl TODO items
- Use SPDX license identifier in README, also do I need to copy the musl license into the folder?
- Try to clear CMake warnings in test, look into the testcase_read parsing failures

Happy for any reviews at this stage.

wolfv:
btw yes, empty `unistd.h` works.

src/repopage.c
```diff
@@ -704,11 +710,22 @@ repopagestore_load_page_range(Repopagestore *store, unsigned int pstart, unsigne
...
+#ifndef _WIN32
           if (pread(store->pagefd, compressed ? buf : dest, in_len, store->file_offset + p->page_offset) != in_len)
 	    {
 	      perror("mapping pread");
 	      return 0;
 	    }
+#else
+	  DWORD read_len;
+	  OVERLAPPED ovlp;
+	  ovlp.Offset = store->file_offset + p->page_offset;
```
mlschroe: Shouldn't you also zero `ovlp.OffsetHigh`?

test/CMakeLists.txt
```diff
@@ -1,12 +1,23 @@
...
-    ENDIF(IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${tcdir}")
-ENDFOREACH(tcdir)
+            IF(IS_DIRECTORY ${dir})
+                FILE(RELATIVE_PATH myname "${CMAKE_CURRENT_SOURCE_DIR}/${tcdir}" ${dir})
+                FILE(TO_CMAKE_PATH "${CMAKE_BINARY_DIR}/tools/testsolv" TESTSOLV_BINARY)
+                FILE(TO_CMAKE_PATH "${dir}" dir)
+                if (WIN32)
+                    STRING(REGEX REPLACE "/" "\\\\" dir ${dir})
+                    STRING(REGEX REPLACE "/" "\\\\" TESTSOLV_BINARY ${TESTSOLV_BINARY})
+                endif ()
```
mlschroe: 1) You deleted my `STRING(REPLACE \\ \\\\ dir ${dir})` line. It is also needed for non-win32 systems.
2) your regex replaces are overkill. You should simply use `FILE(TO_NATIVE_PATH` instead. The \ -> \\ replace will then give you the desired result
3) do you really need the `FILE(TO_CMAKE_PATH` calls? Isn't the `FILE(GLOB` result already a cmake path?

tools/CMakeLists.txt
```diff
@@ -4,7 +4,11 @@
 
 ADD_LIBRARY (toolstuff STATIC common_write.c)
 
-SET (tools_list mergesolv dumpsolv installcheck testsolv repo2solv)
+IF (WIN32)
+    INCLUDE_DIRECTORIES (${PROJECT_SOURCE_DIR}/win32/)
+ENDIF ()
+
+SET (tools_list testsolv)
```
mlschroe: What's with the other tools? `dumpsolv` is pretty useful to have. And the others don't hurt either.

wolfv: I was successful in compiling all other default tools except repo2solv.

Additionally, enabling the `-DDEBIAN` flag now also works on windows.

test/CMakeLists.txt
wolfv: I simplified it, warnings are gone :+1:

src/repopage.c
wolfv: I added an initialization for OVERLAPPED, just like postgres does in pg_pread here: https://github.com/postgres/postgres/blob/74eb2176bf3ac7a2fae1282b4f4f1d000f3e6d78/src/port/pread.c#L29-L48

wolfv:
@mlschroe or someone else, could you take a look at the tests? I see some slight discrepancies:

- Results differ in C:\projects\libsolv\test\testcases\selection
```
Results differ:
15: -job noop pkg A-2-2.i686@available [noautoset]
```

And 

```
      Start 19: weakdeps
19: Test command: C:\projects\libsolv\test\runtestcases.bat "C:\projects\libsolv\tools\testsolv" "C:\projects\libsolv\test\testcases\weakdeps"
19: Test timeout computed to be: 10000000
19: C:\projects\libsolv\test\testcases\weakdeps
19: "C:\projects\libsolv\test\testcases\weakdeps\supplements_implicitobsoleteusescolors.t"
19: test 1:
19: Found 1 problems:
19: Problem 1:
19: package A does not exist
19: 
19: Solution 1:
19:   - do not ask to install A
19: 
19: test 2:
19: Found 1 problems:
19: Problem 1:
19: package A2 does not exist
19: 
19: Solution 1:
19:   - do not ask to install A2
19: 
19: "PASSED";
19/20 Test #19: weakdeps
```


Any ideas? Are these results still correct?

wolfv:
Again the link: https://ci.appveyor.com/project/wolfv/libsolv

win32_compat/README
wolfv: I've added the complete License in the `LICENSE` file in the win32 folder.

win32_compat/README
wolfv: removed all cc-by-sa

CMakeLists.txt
wolfv: fixed.

CMakeLists.txt
wolfv: ext is building on windows now, tests are running and the default tools are available, too.

.appveyor.yml
```diff
@@ -0,0 +1,34 @@
...
+  - ps: if($env:Platform -eq "x86"){Start-FileDownload 'http://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86.exe' C:\Miniconda.exe; echo "Done"}
+  - cmd: C:\Miniconda.exe /S /D=C:\libsolv-conda
+  - "set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%MINICONDA%\\Library\\bin;%PATH%"
+
+install:
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+  - conda install cmake zlib xz -c conda-forge
+  - cmake -G "NMake Makefiles" -D CMAKE_INSTALL_PREFIX=%MINICONDA%\\LIBRARY -DDISABLE_SHARED=1 -DWITHOUT_COOKIEOPEN=1 -DMULTI_SEMANTICS=1 -DDEBIAN=ON -DENABLE_COMPLEX_DEPS=ON .
```
Conan-Kudo: You should probably not set `-DDEBIAN=1` and set `-DENABLE_CONDA=1`. This is not for a Debian target system.

CMakeLists.txt
```diff
@@ -370,10 +370,15 @@ INCLUDE_DIRECTORIES (${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} ${CMAKE_SOU
...
-set (CMAKE_C_FLAGS_DEBUG     "${CMAKE_C_FLAGS} -g3 -O0")
+IF (NOT MSVC)
+  set (CMAKE_C_FLAGS                "${CMAKE_C_FLAGS} -Wall")
+  set (CMAKE_C_FLAGS_RELEASE        "${CMAKE_C_FLAGS} -O3")
+  set (CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS} -g -O3")
+  set (CMAKE_C_FLAGS_DEBUG          "${CMAKE_C_FLAGS} -g3 -O0")
+ELSE ()
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4244 /wd4267")
+  add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+ENDIF ()
```
Conan-Kudo: Can you please flip this to be `IF (MSVC) ... ELSE () ... ENDIF ()`?

CMakeLists.txt
```diff
@@ -413,10 +418,12 @@ ENDIF (HAVE_LINKER_AS_NEEDED)
 ADD_SUBDIRECTORY (src)
 ADD_SUBDIRECTORY (ext)
 ADD_SUBDIRECTORY (tools)
+IF (NOT WIN32)
 IF (ENABLE_PERL OR ENABLE_PYTHON OR ENABLE_RUBY OR ENABLE_TCL)
     ADD_SUBDIRECTORY (bindings)
 ENDIF (ENABLE_PERL OR ENABLE_PYTHON OR ENABLE_RUBY OR ENABLE_TCL)
 ADD_SUBDIRECTORY (examples)
+ENDIF (NOT WIN32)
```
Conan-Kudo: This shouldn't be necessary, since bindings aren't enabled by default anyway. It'd probably be better for examples to get a dedicated configuration switch rather than wiring this into a `NOT WIN32` condition.

ext/testcase.c
```diff
@@ -28,6 +28,14 @@
 #include "ext/repo_helix.h"
 #endif
 
+#ifdef _WIN32
+  #include <direct.h>
+  #define PATH_SEP '\\'
+#else
+  #define PATH_SEP '/'
+#endif
```
Conan-Kudo: I thought `/` works as a path separator natively in Win32 as well?

Conan-Kudo:
@wolfv You'll probably want to rebase onto master before addressing my comments.

test/runtestcases.bat
```diff
@@ -0,0 +1,25 @@
...
+SET cmd=%1
+SET dir=%2
+
+echo %dir%
+
+FOR /f "tokens=*" %%G IN ('dir /b %dir%\*.t') DO (
+  ECHO "%dir%\%%G"
+  CALL %cmd% "%dir%\%%G"
+
+  IF %ERRORLEVEL% EQU 0 (
```
mlschroe: This doesn't seem to work (it always writes PASSED). I think you need to use !ERRORLEVEL! to request delayed expansion.

wolfv: thanks for the remark!

Conan-Kudo:
@wolfv Could you rebase on master?

wolfv:
yep, just wanted to quikcly check the delayed expansion stuff.

wolfv:
Now we seem to successfully have a failure! :)

```
15: "C:\projects\libsolv\test\testcases\selection\selection_matchdeps.t"
15: test 1:
15: test 2:
15: Results differ:
15: -job noop oneof A-2-1.noarch@available A-2-2.i686@available
15: test 3:
15: test 4:
15: test 5:
15: test 6:
15: test 7:
15: Results differ:
15: -job noop pkg A-2-2.i686@available [noautoset]
15: test 8:
15: test 9:
15: "FAILES";
15: "C:\projects\libsolv\test\testcases\selection\selection_matchsolvable.t"
```

Any ideas what the reason could be?

mlschroe:
It fails because musl's fnmatch doesn't support FNM_CASEFOLD...

We might want to switch to some BSD implementation.

mlschroe:
(Or get fnmatch from a newer musl version, they added support for case folding in 2014)

wolfv:
Thanks for looking into this. Will update the fnmatch then :)

.appveyor.yml
wolfv: conda is only used to install dependencies here. I thought debian was necessary for the tests to pass, as that's how the test are compiled on linux.

Conan-Kudo: No, you need to enable _a_ type target, yes, but you don't need to declare a distribution.

ext/testcase.c
wolfv: Does not seem to be the case (https://ci.appveyor.com/project/wolfv/libsolv/builds/23770809) I'll keep it as it is for now if you don't mind.

CMakeLists.txt
wolfv: done

CMakeLists.txt
wolfv: done

.appveyor.yml
wolfv: I removed -DDEBIAN, tests still pass.

wolfv:
Ok I consider this PR in a mergeable state now. I updated the musl files to all be derived from the latest version (1.1.21) and the tests all pass.

I'm happy to get more feedback and will try to fix remaining issues that prevent a merge as quick as possible.

mlschroe:
Here's some more work for you

- please squash some commits (no "update musl")
- I don't want that "ENABLE_EXAMPLE" flags. It should always add the examples dir. Just don't build the "solv" example for MSVC (like you don't build repo2solv)
- why is there this tools_util.h include in testcase.c?

wolfv:
Squashed, removed ENABLE_EXAMPLE (didn't need to change anything since we also removed the debian cmake arg so the examples are not built anyways) nad also fixed up tools_util (that was a rebase gone wrong).

mlschroe:
Thanks a lot for all your work!

wolfv:
Thanks! Really happy to see this merged! And I am also really looking forward to test your conda version matching etc. Haven't gotten around to it yet!

wolfv:
btw you might want to add appveyor support for this repo so that PR's are continually tested as with Travis.

PR #307, opened 2019-03-21
repodata: Avoid crash in repodata_add_dirstr() if dir==0

rluzynski:
repodata_add_dirstr() crashes if called with dir==0 so it should not
be called with that value. Previously it was correctly worked around
in commit 3edcaace14c71804ac571068276fe81880880478, then it was forgotten
in commit 72e8b9af78a98e7226d1d30e3f00848c8b4c36d9. This causes random
crashes, e.g.: https://bugzilla.redhat.com/show_bug.cgi?id=1640278
or https://bugzilla.redhat.com/show_bug.cgi?id=1652017.

### DIFF
--- a/src/repodata.c
+++ b/src/repodata.c
@@ -2941,7 +2941,8 @@ repodata_set_kv(Repodata *data, Id solvid, Id keyname, Id keytype, KeyValue *kv)
         repodata_add_dirnumnum(data, solvid, keyname, kv->id, kv->num, kv->num2);
       break;
     case REPOKEY_TYPE_DIRSTRARRAY:
-      repodata_add_dirstr(data, solvid, keyname, kv->id, kv->str);
+      if (kv->id)
+        repodata_add_dirstr(data, solvid, keyname, kv->id, kv->str);
       break;
     case_CHKSUM_TYPES:
       repodata_set_bin_checksum(data, solvid, keyname, keytype, (const unsigned char *)kv->str);


rluzynski:
I've been experiencing continuous crashes of packagekitd in my Fedora 28 for several months. This not only disabled updates via packagekit-backed frontends but slowed the machine down because packagekitd retried its job every several minutes consuming lots of resources and effectively freezing the machine for few seconds or minutes. Each retry finished with its crash and retry in several minutes… and so on.

I admit this bug is difficult to reproduce and I don't know what triggers it. It was suggested that the reason is a database corruption. But once it happens in happens all the time. However, the analysis of the source code shows the bug is real. Please see the commit 72e8b9af78a98e7226d1d30e3f00848c8b4c36d9 and compare the old version (only relevant excerpts here):

```C
solvable_copy_cb(void *vcbdata, Solvable *r, Repodata *fromdata, Repokey *key, KeyValue *kv)
{
  /* ... */
   switch(key->type)
    {
    /* ... */
    case REPOKEY_TYPE_DIRNUMNUMARRAY:
      id = kv->id;
      id = copydir(pool, data, fromdata, id, cbdata->dircache);
      if (id)    /* ✅ OK, only if id is not zero  */
        repodata_add_dirnumnum(data, handle, keyname, id, kv->num, kv->num2);
      break;
    case REPOKEY_TYPE_DIRSTRARRAY:
      id = kv->id;
      id = copydir(pool, data, fromdata, id, cbdata->dircache);
      if (id)    /* ✅ OK, only if id is not zero  */
        repodata_add_dirstr(data, handle, keyname, id, kv->str);
       break;
  /* ... */
    }
  return 0;
}
```

and the new one:

```C
solvable_copy_cb(void *vcbdata, Solvable *r, Repodata *fromdata, Repokey *key, KeyValue *kv)
{
  /* ... */
  switch (key->type)
    {
      /* ... */
      case REPOKEY_TYPE_DIRNUMNUMARRAY:
      case REPOKEY_TYPE_DIRSTRARRAY:
        kv->id = copydir(data, fromdata, kv->id, cbdata->dircache);
        /* ✅ OK, uses kv->id instead of id  */
        break;
      /* ... */
    }
  repodata_set_kv(data, handle, key->name, key->type, kv);
  return 0;
}

/* ... */

void
repodata_set_kv(Repodata *data, Id solvid, Id keyname, Id keytype, KeyValue *kv)
{
  switch (keytype)
  {
    /* ... */
    case REPOKEY_TYPE_DIRNUMNUMARRAY:
      if (kv->id)
      /* ✅ OK, condition if (kv->id) in place of the old if (id)  */
        repodata_add_dirnumnum(data, solvid, keyname, kv->id, kv->num, kv->num2);
      break;
    case REPOKEY_TYPE_DIRSTRARRAY:
      /* ❌ But where is if (kv->id) in place of the old if (id) ?  */
      repodata_add_dirstr(data, solvid, keyname, kv->id, kv->str);
      break;
```

While debugging the issue once I introduced this fix, installed the corrected version of libsolv and launched packagekitd it completed successfully for the first time in months and offered updates to the GUI frontend. Since then it runs with no problems even if I reverted the changes and installed the original version of libsolv. My guess is that packagekitd managed to rebuild its database or cache or whatever, and the bug does not prevent it from running completely but just prevents it to recover from a previous crash (which may happen at random reasons, consider a power failure or a system shutdown while packagekitd is running in the background).

mlschroe:
The thing is that id == 0 must not happen at all with the current code and points to a problem somewhere else.

Anyway, I'm find with taking this pull request for now. I plan to add support for id zero sometime later, so that src rpms have a correct file list.

rluzynski:
I agree that it means there is a problem somewhere else. Otherwise I would ask to remove the assert. I suggest you to investigate why ```copydir``` returns 0. Maybe it correctly returns 0 to report a minor problem? Maybe the problem is really minor, for example something like _the cache is broken, don't use it, build a new one_?

Simply speaking, assert abort means a programmer's error and it's a bad user's experience to see the crash especially when the software seems to work correctly when the crash is just skipped.

mlschroe:
True. The code will now no longer copy bad packages from the old solv file.

See commit 967a7598adabe9ca5972c7eb08752d49f8bb4c4a

PR #310, opened 2019-03-25
Add support for arch "aarch64:m3ulcb" in poolarch.c

j-mracek:
https://bugzilla.redhat.com/show_bug.cgi?id=1666835

### DIFF
--- a/src/poolarch.c
+++ b/src/poolarch.c
@@ -55,6 +55,7 @@ static const char *archpolicies[] = {
   "sparcv8",	"sparcv8:sparc",
 #if defined(FEDORA) || defined(MAGEIA)
   "ia32e",	"ia32e:x86_64:athlon:i686:i586:i486:i386",
+  "aarch64",	"aarch64:m3ulcb",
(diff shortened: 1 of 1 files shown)


mlschroe:
libsolv follows rpm for the compatible architecture list. Please push this to rpm first.

j-mracek:
Anyway it looks like that the approach is incorrect according to reporter of the original issue.

claneys:
A bit yes, the solution would be to have this mapping someway, somewhere overridable by a file/env var, like RPM does with its 'rpmrc' file or DNF with its '/etc/dnf/vars/arch' file. The solution would be to let the user configure its system instead of adding any weird arch upstream which is definitely not a solution, I agreed.

Does libsolv could rely on the rpmrc file by example to get its arch mapping ?

mlschroe:
libsolv already supports custom arch policies, you can use pool_setarchpolicy() instead of pool_setarch().

j-mracek:
@mlschroe Thank you for perfect hint. We will think how to implement it in libdnf

PR #311, opened 2019-03-28
Add support for computing hashes using OpenSSL

jrohel:
Due to security reasons we would prefer that libsolv would provide hash algorithms based on verified libraries rather then internal implementation.
OpenSSL is often installed on the system why do not used it?
The PR adds OpenSSL support. OpenSSL support is optional. The internal implementation is compiled by default.

There is also an option remove internal implementation. What do you think?


src/chksum.c
```diff
@@ -42,22 +54,46 @@ solv_chksum_create(Id type)
   switch(type)
     {
     case REPOKEY_TYPE_MD5:
+#ifdef WITH_OPENSSL
```
ignatenkobrain: why not to do on top of the file something like `#define solv_MD5_Init MD5_init`?

src/CMakeLists.txt
```diff
@@ -33,16 +32,22 @@ IF (ENABLE_CONDA)
...
+ENDIF (NOT WITH_OPENSSL)
 
 SET (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 IF (HAVE_LINKER_VERSION_SCRIPT)
 SET (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${LINK_FLAGS} -Wl,--version-script=${CMAKE_SOURCE_DIR}/src/libsolv.ver")
 ENDIF (HAVE_LINKER_VERSION_SCRIPT)
 
 IF (DISABLE_SHARED)
-ADD_LIBRARY (libsolv STATIC ${libsolv_SRCS})
+  ADD_LIBRARY (libsolv STATIC ${libsolv_SRCS})
```
ignatenkobrain: this one should be linked with openssl if enabled too (IIRC OpenSSL supports static linking).

src/chksum.c
jrohel: @ignatenkobrain OK I changed it.

src/CMakeLists.txt
jrohel: @ignatenkobrain I done change. But I am not sure if I understand your request.

ignatenkobrain:
I think this looks good. I'm not sure if I agree putting arguments to #define, but tnew version looks cleaner.

I still need to test it though.

jrohel:
@ignatenkobrain
> I'm not sure if I agree putting arguments to #define

Arguments are important. Especialy in the case: `#define solv_SHA1_Final(ctx, md) SHA1_Final(md, ctx)` because the macro changes order of arguments.

But I think that my last change in src/CMakeList.txt are not correct. I think I have to put this back:
```
IF (WITH_OPENSSL)
  TARGET_LINK_LIBRARIES (libsolv ${OPENSSL_CRYPTO_LIBRARY})
ENDIF (WITH_OPENSSL)
```
I can compile libsolv without this but "ldd libsolv.so" does not show dependency on libcrypto, but libsolv.so needs libcrypto.

mlschroe:
Bah, I *hate* the idea of linking to some fat ssl library just for checksum calculation. It's really "taking a sledgehammer to crack a nut". (Do md5sum calculations even work when FIPS is enabled?)

It's almost impossible that the checksum code has a security issue, but that is a moot argument anyway: it's not relevant for dnf/libdnf, as they don't use libsolv for checksum calculation.

So, I really don't want this in libsolv. Please maintain it as a patch for the Fedora libsolv package if you really must do this.

*rant*: It's insane that everybody who wants checksums must either have their own implementations or link against some fat library. Such basic functionality really should be provided by glibc. Actually, glibc already implements all the checksums because it needs them for crypy(). There's just no API function to access them.

Conan-Kudo:
@jrohel I don't think this is strictly necessary for the Fedora/RHEL package either, as these particular checksums are simple and I'm sure SUSE has audited them appropriately (they are already too paranoid for my taste in every other respect...).

jrohel:
@mlschroe 
> Bah, I hate the idea of linking to some fat ssl library just for checksum calculation. It's really "taking a sledgehammer to crack a nut".

I understand your point of view. I don't like dependencies on fat libraries too.
But from another point of view:
- I added **possibility** to use OpenSSL library. **You can still use internal implementation.**
- The libsolv package depends on OpenSSL in Fedora because (librpmio.so depends on libcrypto.so):
```
lddtree /usr/lib64/libsolvext.so
  libsolvext.so => /usr/lib64/libsolvext.so (interpreter => none)
    libsolv.so.1 => /lib64/libsolv.so.1
    librpmio.so.8 => /lib64/librpmio.so.8
        libelf.so.1 => /lib64/libelf.so.1
        libpopt.so.0 => /lib64/libpopt.so.0
        liblua-5.3.so => /lib64/liblua-5.3.so
        libm.so.6 => /lib64/libm.so.6
        libcrypto.so.1.1 => /lib64/libcrypto.so.1.1
        libdl.so.2 => /lib64/libdl.so.2
        libpthread.so.0 => /lib64/libpthread.so.0
        ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
    librpm.so.8 => /lib64/librpm.so.8
        libcap.so.2 => /lib64/libcap.so.2
        libacl.so.1 => /lib64/libacl.so.1
            libattr.so.1 => /lib64/libattr.so.1
        libdb-5.3.so => /lib64/libdb-5.3.so
    libxml2.so.2 => /lib64/libxml2.so.2
    libz.so.1 => /lib64/libz.so.1
    liblzma.so.5 => /lib64/liblzma.so.5
    libbz2.so.1 => /lib64/libbz2.so.1
    libzstd.so.1 => /lib64/libzstd.so.1
    libzck.so.1 => /lib64/libzck.so.1
    libc.so.6 => /lib64/libc.so.6
```
So there is allways OpenSSL installed on Fedora system and user is forced to use internal implementation in libsolv.

> Do md5sum calculations even work when FIPS is enabled?

Which operations in libsolv need MD5? Is it permitted use MD5 for these operations on FIPS systems?

>It's almost impossible that the checksum code has a security issue, but that is a moot argument anyway: it's not relevant for dnf/libdnf, as they don't use libsolv for checksum calculation.

Libsolv provides `solv_chksum_*` functions. They can be used by anyone to anything. But you are right we shouldn't use them. 

> So, I really don't want this in libsolv. Please maintain it as a patch for the Fedora libsolv package if you really must do this.

I don't like this. But if there is not another way.

> rant: It's insane that everybody who wants checksums must either have their own implementations or link against some fat library. Such basic functionality really should be provided by glibc. Actually, glibc already implements all the checksums because it needs them for crypy(). There's just no API function to access them.

Yes it's insane.

ignatenkobrain:
> So there is allways OpenSSL installed on Fedora system and user is forced to use internal implementation in libsolv.

Strictly speaking, it is only libsolvext while chksum interface is part of libsolv.

jrohel:
@ignatenkobrain
> Strictly speaking, it is only libsolvext while chksum interface is part of libsolv.

Yes, I know. But both libraries (libsov and libsolvext) are provided by the same package libsolv. So, user have no choice to install only libsolv (without libsolvext) on Fedora. As I wrote above "The libsolv **package** depends on OpenSSL in Fedora because...".

mlschroe:
You're somewhat missing my point: it does not matter if it's an option and what the default is. When I accept the pull request I have to maintain the code and react to bug reports.

Regarding md5: it's used in the testcase code to create unique identifiers for problems, rules and the like. But this is not an operation which needs a cryptographic hash function, so md5 is perfectly find in that context.

Issue #312, opened 2019-04-02
0.7.3/0.7.4 regression in SOLVER_FAVOR handling


hanzz:
Hi,

Fedora just updated to libsolv 0.7.4 and our tests for module-build-service started failing. I'm attaching solver test describing the issue:

[solv-0.7.4-regression.tar.gz](https://github.com/openSUSE/libsolv/files/3033350/solv-0.7.4-regression.tar.gz)

We favored gtk:2:0:c4-0.x86_64 solvable in a test, but the solver chose gtk:1:0:c2-0.x86_64 despite the fact that gtk:2:0:c4-0.x86_64 solvable has also satisfied dependencies.

This test worked as expected (chose gtk:2:0:c4-0.x86_64) in the older libsolv-0.7.2.

Can you please verify if that's regression in libsolv? Or do we need to change our code somehow?

Updated: I cannot reproduce it using "testsolv" and I can reproduce it only using python3.

hanzz:
I also think this might be related commit, but haven't tried doing bisect to verify that: https://github.com/openSUSE/libsolv/commit/2645bc64b322460ac77d0d051de63434d4a90a0a

ignatenkobrain:
```
⋊> ~/D/s/test testsolv testcase.t                                                                                                                                           12:05:11
Results differ:
-install gtk:1:0:c2-0.x86_64@available
+install gtk:2:0:c4-0.x86_64@available
⋊> ~/D/s/test rpm -q libsolv                                                                                                                                                12:05:15
libsolv-0.7.4-1.fc31.x86_64
```

hanzz:
I'm using libsolv-0.7.4-1.fc29.x86_64.

hanzz:
Hm, and I also cannot reproduce it using testsolv :(...

```
$ testsolv testcase.t 
Results differ:
-install gtk:1:0:c2-0.x86_64@available
+install gtk:2:0:c4-0.x86_64@available
```

hanzz:
I also found out that in MBS test-suite, I can only reproduce that on python3.

mlschroe:
I can't reproduce the problem here. If I run the testcase the solver correctly installs gtk:2:0:c4-0.x86_64@available. Can you please attach the output of 'testsolv -v -v -v testcase.t'?

mlschroe:
(Ah, you already found that out yourself. I should have reloaded that github page)

mlschroe:
[...]

PR #313, opened 2019-04-02
solver: Free favorq when running solver_solve()

ignatenkobrain:
Otherwise, favorq leaks from previous executions.

Fixes: https://github.com/openSUSE/libsolv/issues/312


mlschroe:
The pull request is certainly correct, but I wonder why that problem didn't show up before. The favorq cleanup code was missing since we introduced favor jobs...

ignatenkobrain:
@mlschroe no, it appeared only after refactoring in https://github.com/openSUSE/libsolv/commit/2645bc64b322460ac77d0d051de63434d4a90a0a.

`isdisfavormap` was always reset before and it was somehow cleaning favormap. Now we have `havedisfavored` which was not reset.

PR #317, opened 2019-04-14
[DO NOT MERGE] add some experiments to test conda solving

wolfv:
Hi @mlschroe 

finally I got around to test some of the shiny new conda support in this repo! :)

To run this new tool, one needs to download a couple of repodata.json files. Here is my current set:

```
wget https://conda.anaconda.org/conda-forge/noarch/repodata.json -O cf.noarch.json
wget https://conda.anaconda.org/conda-forge/linux-64/repodata.json -O cf.linux64.json
wget https://conda.anaconda.org/anaconda/linux-64/repodata.json -O anaconda.linux64.json
wget https://conda.anaconda.org/anaconda/noarch/repodata.json -O anaconda.noarch.json
```

and put these files into a dedicated folder (e.g. "repos")

Then 

`condatest -d ./repos -i bqplot`

solves for bqplot! 

So far I've noticed a couple of issues. For example, the above command tries to install bqplot, but pulls in Python 2.7 as dependency even though alternatively Python 3.7 is available. In the past I've been able to debug these kinds of issues by adding another solving objective and looking for errors 

`condatest -d ./repos -i bqplot -i "python 3.7"`

but this also runs fine! 

I was wondering what's the best way to debug this? 

Cheers!

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -104,9 +104,11 @@ ENDIF (ENABLE_APPDATA)
 
 IF (ENABLE_CONDA)
 ADD_EXECUTABLE (conda2solv conda2solv.c)
+ADD_EXECUTABLE (condatest condatest.c)
 TARGET_LINK_LIBRARIES (conda2solv toolstuff libsolvext libsolv ${SYSTEM_LIBRARIES})
+TARGET_LINK_LIBRARIES (condatest toolstuff libsolvext libsolv ${SYSTEM_LIBRARIES})
 
-SET (tools_list ${tools_list} conda2solv)
+SET (tools_list ${tools_list} conda2solv condatest)
 ENDIF (ENABLE_CONDA)
 
 IF (NOT WIN32)
--- a/tools/condatest.c
+++ b/tools/condatest.c
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2019, SUSE LLC
+ *
+ * This program is licensed under the BSD license, read LICENSE.BSD
+ * for further information
+ */
+
+/*
+ * conda2solv.c
+ *
+ * parse a conda repository file
+ *
+ * reads from stdin
+ * writes to stdout
+ */
+
+#include <dirent.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "common_write.h"
+#include "conda.h"
+#include "pool.h"
+#include "repo.h"
+#include "repo_conda.h"
+#include "solv_xfopen.h"
+#include "solverdebug.h"
+
+static void usage(int status) {
+  fprintf(stderr, "\nUsage:\n"
+                  "conda2solv\n"
+                  "  reads a 'synthesis' repository from <stdin> and writes a "
+                  ".solv file to <stdout>\n"
+                  "  -h : print help & exit\n");
+  exit(status);
+}
+
+int main(int argc, char **argv) {
+  Pool *pool;
+  Repo *repo;
+  int c;
+  DIR *d;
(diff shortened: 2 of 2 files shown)


wolfv:
Interestingly the "mamba" prototype with much more hackier version comparison etc. seems to not suffer from this problem.

wolfv:
Another observation is that when I add two jobs to the solver, one `python` and another pure python package (I randomly chose `decorator`) the ordering of the two influences which python version get's chosen:

`./tools/condatest -d ./repos -i python -i decorator` 
installs python 3.7 and decorator 4.4

`./tools/condatest -d ./repos -i python -i decorator` 
installs python 2.7 and decorator 4.4

I think it's also OK to not consider this a bug and assume that the python version should be selected before attempting to solve as an environment wide spec ... but maybe we can also find a way to maximize the python version number?

mlschroe:
I'm actually surprised that it does anything at all, because you use `SOLVER_INSTALL | REL_CONDA` which makes no sense at all. REL_CONDA is for constructing dependencies, you can't use it in a job. You should use `SOLVER_INSTALL | SOLVER_SOLVABLE_PROVIDES`.

(But the solver seems to fallback to this SOLVABLE_PROVIDES unknown selection)

As for how to debug this, you need to set the debuglevel to 2 or 3. Then you'll see that the solver actually tried to install python-3.7.3-h5b0a415_0 first but ran into a problem.

It's probably best to change the code so that it writes a solver testcase with testcase_write().
(Actually you could simply use testsolv instead of testconda for testing various installation jobs.)

So, what's going on with python3? The problem is that it pulled in pandas-0.24.2-py27hf484d3e_0, which conflicts with python3. Basically it picked some python version pretty early. The first mention of py27 is when it decides which numpy to use. After version pruning it has:
- numpy-1.16.2-py27_blas_openblash1522bff_0 [38272]
- numpy-1.16.2-py27h7fc5cc3_0 [38273]
- numpy-1.16.2-py27h8b7e671_1 [38274]
- numpy-1.16.2-py36_blas_openblash1522bff_0 [38275]
- numpy-1.16.2-py36h7fc5cc3_0 [38276]
- numpy-1.16.2-py36h8b7e671_1 [38277]
- numpy-1.16.2-py37_blas_openblash1522bff_0 [38278]
- numpy-1.16.2-py37h7fc5cc3_0 [38279]
- numpy-1.16.2-py37h8b7e671_1 [38280]
- numpy-1.16.2-py27h7e9f1db_0 [85651]
- numpy-1.16.2-py27h99e49ec_0 [85652]
- numpy-1.16.2-py36h7e9f1db_0 [85653]
- numpy-1.16.2-py36h99e49ec_0 [85654]
- numpy-1.16.2-py37h7e9f1db_0 [85655]
- numpy-1.16.2-py37h99e49ec_0 [85656]

Of those it picks the first one.

mlschroe:
The underlying problem is IMHO that libsolv's prune_to_best_version() function does not understand build flavors yet.

mlschroe:
(Btw, I think that you got different results with the mambo prototypo because it ignored the buildflavor part of the package dependencies. So I wouldn't trust the result from mamba.)

wolfv:
I think it "works" in mamba because the version is something like `0.1.3.0-py37` which probably is resolved as higher than `0.1.3.0-py27`.

mlschroe:
That could also be. It'll also work similar like this with my next commit, which will also sort by build flavor. Sorting by build flavor is good anyway, because it removes some bit of randomness. Still, we might want to handle this in a more sane way in the future.

wolfv:
I can confirm that with the latest changes it properly also "maximizes" the build string! cool!

wolfv:
Closing, since I am happy to test stuff in mamba itself now! Thanks again for the effort, it's working really well so far!

PR #318, opened 2019-04-16
add star even if build string is present

wolfv:
In order to install a definition such as this one:

`python =3.7 h381d211_0`

I needed to remove this condition so that this gets translated to 

`python 3.7* h381d211_0`

I'm not sure what purpose the condition on the build string was serving. Let me know if this needs a different kind of fix.

### DIFF
--- a/src/conda.c
+++ b/src/conda.c
@@ -645,8 +645,6 @@ pool_conda_matchspec(Pool *pool, const char *name)
 	  if (!build)
 	    version += 2;
 	}
-      else if (build)
-	version += 1;
       else
 	{
 	  for (p = version + 1; p < versionend; p++)


mlschroe:
The condition is there because that's what _parse_spec_str() in conda's match_spec.py does.
If I call it with `python 3.7* h381d211_0` I get the following:
```
{'version': '3.7', 'name': 'python', 'build': 'h381d211_0'}
```
The version does not have a trailing '*'.

wolfv:
Ok, I see, that makes sense. 
For the first variant I get a "unsupported request" problem shown, while the second one works as expected.
I guess I should look into how solvalbles get matched/found then to make sure that `3.7` without other specifiers matches all `3.7*` versions.

mlschroe:
Well, you need to convince me that it works in conda first ;)

mlschroe:
Because conda's get_matcher() function in version.py makes this an exact match. So there must be something I'm missing, because conda's code makes it look like `python =3.7 h381d211_0` will only match an exact 3.7 version.

wolfv:
I think you're right! Give me a second to see where I went wrong.

wolfv:
Hmm, weird, you're definitely correct. I don't even know why I tried to solve this :) I guess I just thought that this was what conda does, but it's definitely not. Sorry!

PR #319, opened 2019-04-16
fix up json parsing and comparison of build version

wolfv:
The build version was never parsed from the json (as it's a JP_NUMBER type) and consequently, the sorting after build version did not work. 

This was evident when trying to install `ncurses` which should pull in ncurses 6.1 1002 but chose ncurses 6.1 0

This seems to fix it.

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -58,8 +58,8 @@ parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn)
     {
       if (type == JP_STRING && !strcmp(jp->key, "build"))
 	repodata_add_poolstr_array(data, handle, SOLVABLE_BUILDFLAVOR, jp->value);
-      else if (type == JP_STRING && !strcmp(jp->key, "build_number"))
-	repodata_set_str(data, handle, SOLVABLE_BUILDVERSION, jp->value);
+      else if (type == JP_NUMBER && !strcmp(jp->key, "build_number"))
+	repodata_set_num(data, handle, SOLVABLE_BUILDVERSION, strtoull(jp->value, 0, 10));
       else if (type == JP_ARRAY && !strcmp(jp->key, "depends"))
 	type = parse_deps(pd, jp, &s->requires);
       else if (type == JP_ARRAY && !strcmp(jp->key, "requires"))
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -31,6 +31,7 @@ SET (libsolv_HEADERS
 
 IF (ENABLE_CONDA)
     SET (libsolv_SRCS ${libsolv_SRCS} conda.c)
+    SET (libsolv_HEADERS ${libsolv_HEADERS} conda.h)
 ENDIF (ENABLE_CONDA)
 
 IF (NOT MSVC)
--- a/src/policy.c
+++ b/src/policy.c
@@ -834,14 +834,9 @@ move_installed_to_front(Pool *pool, Queue *plist)
 static int
 pool_buildversioncmp(Pool *pool, Solvable *s1, Solvable *s2)
 {
-  const char *bv2, *bv1 = solvable_lookup_str(s1, SOLVABLE_BUILDVERSION);
-  if (bv1)
-    {
-      bv2 = solvable_lookup_str(s2, SOLVABLE_BUILDVERSION);
-      if (bv1 != bv2)
-	return pool_evrcmp_str(pool, bv1, bv2, EVRCMP_COMPARE);
-    }
-  return 0;
+  const unsigned long bv1 = solvable_lookup_num(s1, SOLVABLE_BUILDVERSION, 0);
+  const unsigned long bv2 = solvable_lookup_num(s2, SOLVABLE_BUILDVERSION, 0);
+  return (bv1 > bv2) - (bv1 < bv2);
 }
 
 static int


wolfv:
but the comparison is also wrong as far as I can tell. If one of the two has a build version and the other doesn't, then it returns 0.

wolfv:
Also this evr cmp basically calls into the full conda version comparison if I see that correctly, and it seems quite wasteful for what should be just a integer comparison (the mode setting is also seemingly ignored). Don't you think we can find a more elegant solution?

Btw. after this, the remaining problem is that I think the build number / build string sorting is taking precedence over repo priority. I think ideally, there should probably be a cascade of sorting ops taking place:

- first repo priority
- then build number (grouped by repo prio)
- then build string (grouped by build number)

It's probably not trivial to implement but if you agree I could take a look at it.

mlschroe:
No, repo prio comes first in libsolv (see policy.c).

mlschroe:
See pool_best_solvables() in policy.c:

```
  if (plist->count > 1) 
    prune_to_highest_prio(pool, plist);
  if (plist->count > 1) 
    prune_to_best_arch(pool, plist);
  if (plist->count > 1) 
    prune_to_best_version(pool, plist);
```

wolfv:
I see. I think I got one "reproducible" example. When trying to install the "munch" library, I would like libsolv to choose the conda-forge version. For this I set the subpriority to somethign higher than the default channel (because I would want to take a higher version from default if available, otherwise prefer conda-forge).
But hte buidl strings differ, e.g.:

```
  - job Rule #42823:
    munch-2.0.4-py27_0 [6434] (w1)
    munch-2.0.4-py34_0 [6435] (w2)
    munch-2.0.4-py35_0 [6436]
    munch-2.0.4-py36_0 [6437]
    munch-2.1.0-py27_0 [6438]
    munch-2.1.0-py34_0 [6439]
    munch-2.1.0-py35_0 [6440]
    munch-2.1.0-py36_0 [6441]
    munch-2.1.1-py27_0 [6442]
    munch-2.1.1-py35_0 [6443]
    munch-2.1.1-py36_0 [6444]
    munch-2.2.0-py27_0 [6445]
    munch-2.2.0-py35_0 [6446]
    munch-2.2.0-py36_0 [6447]
    munch-2.3.0-py_0 [23318]
    munch-2.3.1-py_0 [23319]
    munch-2.3.2-py_0 [23320]
    munch-2.3.1-py27_0 [34261]
    munch-2.3.1-py35_0 [34262]
    munch-2.3.1-py36_0 [34263]
    munch-2.3.2-py27_0 [34264]
    munch-2.3.2-py35_0 [34265]
    munch-2.3.2-py36_0 [34266]
    munch-2.3.2-py37_0 [34267]
    munch-2.0.4-py27_0 [68099]
    munch-2.0.4-py34_0 [68100]
    munch-2.0.4-py35_0 [68101]
    munch-2.0.4-py36_0 [68102]
    munch-2.1.0-py27_0 [68103]
    munch-2.1.0-py35_0 [68104]
    munch-2.1.0-py36_0 [68105]
    munch-2.1.1-py27_0 [68106]
    munch-2.1.1-py35_0 [68107]
    munch-2.1.1-py36_0 [68108]
    munch-2.2.0-py27_0 [68109]
    munch-2.2.0-py35_0 [68110]
    munch-2.2.0-py36_0 [68111]
    next rules: 0 0
```

where we see condaf forge has `munch-2.3.2-py_0` and conda default has 

```
    munch-2.3.2-py27_0 [34264]
    munch-2.3.2-py35_0 [34265]
    munch-2.3.2-py36_0 [34266]
    munch-2.3.2-py37_0 [34267]
```

Now conda selects the conda forge `py0` version here, while libsolv currently selects the default channel `py37_0` version. Maybe you have a good idea how to fix this?

Cheers!

mlschroe:
Hmm, I first need to understand why conda selects the py0 package. How do you configure that is should prefer the conda forge packages?

wolfv:
Here is actually a good description:

https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#after-conda-4-1-0

I have a sorted list of channels with conda-forge at the top. In the conda test script I made it so that if the filename of the repo begins with a digit, that is set to the repo subpriority.

However I am now doubting if the mapping to subpriority is correct ... after reading the (default) conda sorting rules. Maybe mapping to priority would be more appropriate.

wolfv:
I think this doc is slightly outdated. Current conda added a "flexible" mode:

```
(base)  wolfv@x86_64-conda_cos6-linux-gnu  ~/Programs/mamba   nextgen ●  conda config --describe channel_priority
# # channel_priority (ChannelPriority)
# #   Accepts values of 'strict', 'flexible', and 'disabled'. The default
# #   value is 'flexible'. With strict channel priority, packages in lower
# #   priority channels are not considered if a package with the same name
# #   appears in a higher priority channel. With flexible channel priority,
# #   the solver may reach into lower priority channels to fulfill
# #   dependencies, rather than raising an unsatisfiable error. With channel
# #   priority disabled, package version takes precedence, and the
# #   configured priority of channels is used only to break ties. In
# #   previous versions of conda, this parameter was configured as either
# #   True or False. True is now an alias to 'flexible'.
# # 
# channel_priority: flexible
```

wolfv:
I was mapping `strict` mode to `repo->priority` and `flexible` to `repo->subpriority`.

mlschroe:
It seems like "flexible" should be mapped to repo->priority wheres "strict" is currently not supported in libsolv (nobody asked for it yet. yum used to have "strict" prios as well, but they seem to be happy with new new "flexible" prios.)

wolfv:
You seem to be right. I'll try that for now!

PR #320, opened 2019-04-18
add repo subprio handling for conda

wolfv:
I came up with this fix for the repo subpriority sorting for conda packages. Not sure this is the correct fix though!


wolfv:
Note after testing on a bigger installation i am not convinced that this does what I am aiming for. also we had some more discussion on https://github.com/openSUSE/libsolv/pull/319#issuecomment-484424282

wolfv:
Closing since using `priority` instead of subpriority does the job.

PR #323, opened 2019-04-19
OS X rpath handling

wolfv:
This turns on rpath support for OS X and removes the (old?) INSTALL_NAME_DIR method.

From cmake --help-policy CMP0042

cmake --help-policy CMP0042
CMP0042
-------

``MACOSX_RPATH`` is enabled by default.

CMake 2.8.12 and newer has support for using ``@rpath`` in a target's install
name.  This was enabled by setting the target property
``MACOSX_RPATH``.  The ``@rpath`` in an install name is a more
flexible and powerful mechanism than ``@executable_path`` or ``@loader_path``
for locating shared libraries.

CMake 3.0 and later prefer this property to be ON by default.  Projects
wanting ``@rpath`` in a target's install name may remove any setting of
the ``INSTALL_NAME_DIR`` and ``CMAKE_INSTALL_NAME_DIR``
variables.

This policy was introduced in CMake version 3.0.  CMake version
3.12.3 warns when the policy is not set and uses OLD behavior.  Use
the cmake_policy command to set it to OLD or NEW explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.


wolfv:
@mlschroe any reason why this wasn't merged?

mlschroe:
No, except having not much time. I'll merge it this week.

mlschroe:
Hmm, I don't think this is correct. We still need INSTALL_NAME_DIR for non-OS X systems.

How about just adding this to the top CMakeLists.txt file:
```
IF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   SET (CMAKE_MACOSX_RPATH ON)
ENDIF ()
```

wolfv:
Are you sure? It's a OS X only property (where the installation is broken for libsolv as far as I am aware)

https://cmake.org/cmake/help/v3.0/prop_tgt/INSTALL_NAME_DIR.html

mlschroe:
Hmm, could be. When I wrote the above comment I looked at the cmake code and there's not much indication that it's macos only. But there's an test if the CMAKE_PLATFORM_HAS_INSTALLNAME property is set, which seems to be only done on macos.

So how about we delete the INSTALL_NAME_DIR properties and add
```
SET (CMAKE_MACOSX_RPATH ON)
```
to the top CMakeLists file? (I don't think we need the `IF` I proposed in my old comment.)

wolfv:
we can also not set it and just remove the INSTALL_NAME_DIR property and then it will pick up the default behavior.

but I am also happy to be explicit and use your suggestion.

wolfv:
modified it as discussed.

wolfv:
btw libsolv has the `appveyor.yml` file in the repo, but the appveyor CI hooks have not been set up. Maybe that could be done? Could be useful to spot regressions for MSVC support early.

PR #336, opened 2019-07-10
Fix lto-bytecode errors:

marxin:
[  250s] libsolv-devel.x86_64: E: lto-bytecode (Badness: 10000) /usr/lib64/libsolv.a
[  250s] libsolv-devel.x86_64: E: lto-bytecode (Badness: 10000) /usr/lib64/libsolvext.a

- Add -ffat-lto-objects to $optflags as the package provides
  static libraries
- Remove NO_BRP_STRIP_DEBUG=true as brp-15-strip-debug will
  not strip debug info for archives

### DIFF
--- a/package/libsolv.changes
+++ b/package/libsolv.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jul 10 07:48:10 UTC 2019 - Martin Liška <mliska@suse.cz>
+
+- Add -ffat-lto-objects to $optflags as the package provides
+  static libraries
+- Remove NO_BRP_STRIP_DEBUG=true as brp-15-strip-debug will
+  not strip debug info for archives
+
+-------------------------------------------------------------------
 Thu Jun 13 16:15:39 CEST 2019 - mls@suse.de
 
 - make cleandeps jobs on patterns work [bnc#1137977]
--- a/package/libsolv.spec.in
+++ b/package/libsolv.spec.in
@@ -207,6 +207,7 @@ Perl bindings for libsolv.
 %setup -q
 
 %build
+%global _lto_cflags %{_lto_cflags} -ffat-lto-objects
 export CFLAGS="%{optflags}"
 export CXXFLAGS="$CFLAGS"
 
@@ -256,10 +257,6 @@ ln -s repo2solv %{buildroot}/%{_bindir}/repo2solv.sh
 %if %{with python3}
 %py3_compile %{buildroot}/%{python3_sitearch}
 %endif
-%endif
-%if %{with static}
-# we want to leave the .a file untouched
-export NO_BRP_STRIP_DEBUG=true
 %endif
 
 %check


marxin:
The problem can be seen here:
https://build.opensuse.org/build/openSUSE:Factory/standard/x86_64/libsolv/_log

mlschroe:
Do you happen to know when brp-15-strip-debug was fixed? It used to stip archives.

mlschroe:
Looking at some old brp-15-strip-debug it does this for archives:
```
# Don't strip debuginfo from static libs, but compiler-generated local symbols
while read f; do
        case $(file "$f") in
            *"current ar"*|*ELF*", not stripped"*)
                chmod u+w "$f" || :
                strip -p --discard-locals -R .comment -R .note "$f" || :
                ;;
            *)
                echo "WARNING: Strange looking archive $(file $f)"
                continue
                ;;
        esac
done < <(find "$RPM_BUILD_ROOT" "${FIND_IGNORE[@]}" -type f -name "*.a" -print)
```

I think this is why the NO_BRP_STRIP_DEBUG was added in 2008. (Not by me, so I can't tell you why this broke things...)

mlschroe:
(This is because libzypp then includes all of libsolv.a and we want the complete debug information in the libzypp debuginfo package)

marxin:
> (This is because libzypp then includes all of libsolv.a and we want the complete debug information in the libzypp debuginfo package)

Yes, it probably used to make sense to set `NO_BRP_STRIP_DEBUG`. Now it's not needed any longer.
Can you please merge the PR?

mlschroe:
Not so fast, what about the "compiler-generated local symbols"?

marxin:
> Not so fast, what about the "compiler-generated local symbols"?

These do not affect the quality of debug info.

mlschroe:
Then why are they in the objects at all?

marxin:
> Then why are they in the objects at all?

That are probably labels generated by compiler for various purposes. Note that I'm not fully familiar with what exactly the option removes (`--discard-locals`). But as said, I haven't seen an issue that would complain about problematic debug info of static libraries.

mlschroe:
What I'm still not getting is the point of removing the BRP_NO_STRIP_DEBUG line? What exactly do you want to fix with this?

marxin:
> What I'm still not getting is the point of removing the BRP_NO_STRIP_DEBUG line? What exactly do you want to fix with this?

Because I need to strip LTO bytecode from the static archives. It's added here:
https://github.com/openSUSE/brp-check-suse/commit/efb1abd6504187a91a9896fcf4b46310105c03fc

mlschroe:
Ok, now I'm curious. The strip only strips the slim lto part, but leaves the fat (gimple) part in place? Is there no option to generate the fat part without the slim part?

marxin:
> Ok, now I'm curious. The strip only strips the slim lto part, but leaves the fat (gimple) part in place?

No, it will remove GIMPLE bytecode and assembly will remain. Which is basically equivalent to non-LTO mode.

> Is there no option to generate the fat part without the slim part?

mlschroe:
So shouldn't we just build without lto, i.e. add:
```
%define _lto_cflags %{nil}
```

mlschroe:
(Don't get me wrong, I don't mind merging your pull request. I'm just a bit cautious because libzypp takes the objects from the libsolv archive and builds a shared library of it, which is a somewhat unusual approach.)

marxin:
> So shouldn't we just build without lto, i.e. add:
> 
> ```
> %define _lto_cflags %{nil}
> ```

No, because we want to have executables and shared libraries being optimized with LTO.

marxin:
> (Don't get me wrong, I don't mind merging your pull request. I'm just a bit cautious because libzypp takes the objects from the libsolv archive and builds a shared library of it, which is a somewhat unusual approach.)

This is fine, an archive (static library) is just a collection of `.o` files and adding then to linking a shared library is very common thing.

marxin:
Thanks. Can you please help me how `https://build.opensuse.org/package/show/zypp:Head/libsolv` package is updated? Will it be done automatically?

mlschroe:
Yes, that's updated automagically via Jenkins.

PR #344, opened 2019-08-06
repodata_schema2id: fix heap-buffer-overflow in memcmp

haoren3696:
When the length of last schema in data->schemadata is
less than length of input schema, we got a read overflow
in asan test.

Signed-off-by: Zhipeng Xie <xiezhipeng1@huawei.com>


mlschroe:
Yes, the code assumed that the buffer is tested from left to right, but this is indeed not a valid assumption.

PR #346, opened 2019-08-08
bindings: fix object reference for swig 4

kkaempf:
This fixes the build against swig 4.0.0


ignatenkobrain:
Should not it be `$self` everywhere? I am asking because of #347

kkaempf:
Well, probably. I just made this change because `solv.i` already had a similar change at lines 632ff.
If #347 fixes the build with swig 4.0, you can discard this pull request.

mlschroe:
Discarding...

PR #347, opened 2019-08-12
Make code compatible with swig 4.0, remove obj0 instances

StefanBruens:
The deprecated "-aliasobj0" option was removed in swig 4.0, and
obj0, which was only meant as a porting helper, is no longer available:
https://github.com/swig/swig/commit/cd8fc0a025fb
Use the correct "$self" notation, which is expanded to the correct
code regardless of swig version and options.


StefanBruens:
See https://build.opensuse.org/package/show/home:StefanBruens:branches:zypp:Head/libsolv

PR #353, opened 2019-10-09
Added creation of packages

KOLANICH:


ignatenkobrain:
Why?

Conan-Kudo:
This is not acceptable. It breaks building libsolv for too many supported targets (RHEL, SLE, etc.), and there's no value in this.

If you *really* want Debian packages for libsolv, we can extend the spec to support it and have it built somewhere.

mlschroe:
At some time in the past when libsolv was called libsatsolver, we already had cpack support. We threw it out because it didn't make much sense -- it's much easier to use github's release mechanism to provide compressed tar balls.

PR #356, opened 2019-12-02
Fixed python discovery using more modern CMake functions.

KOLANICH:
For old CMake they are bundled and backported. For newer CMake the ones shipped with CMake are used.


KOLANICH:
>Why do we want to ship Python 2 stuff at all anymore?

I am happy to drop it. I have not been using python 2 since 2013. But since python 2 support is already in the repo, I have implemented it too.

KOLANICH:
@mlschroe, @ignatenkobrain

mlschroe:
Where's the "fixed discovery"? The commit is just doing things different, but I don't see what it tries to fix.

KOLANICH:
>Where's the "fixed discovery"? The commit is just doing things different, but I don't see what it tries to fix.

Now I don't remember all the details what the issues were, but was is related to python 2 being used somewhere where python 3 must have been used. Instead of fixing your discovery code I have just replaced it with CMake one and that has worked. Since the lib supports old versions of CMake without these modules, I have just backported and bundled them (for newer CMake the modules shipped with CMake are used). When support of old CMake versions will be dropped these backported modules can be removed.

PR #357, opened 2019-12-02
Fixed discovery of LZMA libraries

KOLANICH:


CMakeLists.txt
```diff
@@ -166,8 +166,8 @@ INCLUDE_DIRECTORIES (${ZLIB_INCLUDE_DIRS})
 ENDIF (ENABLE_ZLIB_COMPRESSION)
 
 IF (ENABLE_LZMA_COMPRESSION)
-FIND_PACKAGE (LZMA REQUIRED)
-INCLUDE_DIRECTORIES (${LZMA_INCLUDE_DIR})
+FIND_PACKAGE (LibLZMA REQUIRED)
+INCLUDE_DIRECTORIES (${LIBLZMA_INCLUDE_DIRS})
```
Conan-Kudo: The `LZMA` "package" comes from the bundled `FindLZMA.cmake` that is in this repository.

KOLANICH: FindLibLZMA is [a part of CMake](https://cmake.org/cmake/help/v3.0/module/FindLibLZMA.html) and is better because it populates info about the version. It is proposed to drop an own wheel and use the one shipped with CMake.

KOLANICH:
@mlschroe, @ignatenkobrain

PR #362, opened 2020-01-15
add parsing of the packages.conda key

wolfv:
conda introduced a new package format (called `.conda`).

In order to stay backward compatible (older versions of conda do not support un-archiving this new package format) they introduced a new key (`packages.conda`).

This PR just adds 4 lines of code to additionally parse the packages found in this key. 

One thing I am not sure of is what will happen if a package is found in _both_, the `packages` and `packages.conda` key. I am currently looking for clarification from the conda folks if we can expect that to never happen.

If this happens, how could we drop the old format packages?

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -163,6 +163,10 @@ parse_main(struct parsedata *pd, struct solv_jsonparser *jp)
 	type = parse_packages(pd, jp);
       if (type == JP_ARRAY && !strcmp("packages", jp->key))
 	type = parse_packages2(pd, jp);
+      if (type == JP_OBJECT && !strcmp("packages.conda", jp->key))
+	type = parse_packages(pd, jp);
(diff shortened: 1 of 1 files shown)


wolfv:
I think -- also because of changes on how mamba works -- we have to make sure to overwrite the `.tar.bz2` packages with the `.conda` packages.

I think most of the time this works because the `package.conda` key will come after the `packages` key, but if that doesn't hold true this might fail.

Do you have some input @mlschroe ?

mlschroe:
We need to do the same thing as conda (_process_raw_repodata_str), i.e. use only the .conda package if a package is available in both sets.

As we have a stream based parser this means replacing the old solvable if the legacy package came first.

mlschroe:
Ok, implemented with commit b582c6588e68b9a161a45a6b94b4feb82843b9d6

wolfv:
cool! do you know when you could make a new release?

PR #363, opened 2020-01-15
de-prioritize track_features (conda)

wolfv:
Conda has a ancient key called `track_features` which was previously used to decide between different "features sets". However, this usage was dropped as it was making the SAT problem too hard & slow (mutex packages are used to achieve a similar effect nowadays).

However, there is one property of the "track_features" key – that it can be used to _deprioritize_ packages, and that is still being used today (cf. https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#track-features )

So this PR attempts to read the `track_features` key if it is avaible in the package, and stores a number in the repo/pool. In the comparison step, depending on how many track_features have been found, the package get's de-prioritized.

Note: the track_features key has a value that is a string, which can contain multiple track features seperated by either a space or `,`.

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -107,6 +107,13 @@ parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn)
 	    ts /= 1000;
 	  repodata_set_num(data, handle, SOLVABLE_BUILDTIME, ts);
 	}
+      else if (type == JP_STRING && !strcmp(jp->key, "track_features"))
+	{
+	  unsigned long long count = 1;
+	  for (unsigned long long i=0; jp->value[i]; i++)
+	    count += (jp->value[i] == ',' || jp->value[i] == ' ');
+	  repodata_set_num(data, handle, SOLVABLE_TRACK_FEATURES, count);
+	}
       else
 	type = jsonparser_skip(jp, type);
     }
--- a/src/knownid.h
+++ b/src/knownid.h
@@ -266,6 +266,7 @@ KNOWNID(UPDATE_STATUS,			"update:status"),		/* "stable", "testing", ...*/
 KNOWNID(LIBSOLV_SELF_DESTRUCT_PKG,      "libsolv-self-destruct-pkg()"),	/* this package will self-destruct on installation */
 
 KNOWNID(SOLVABLE_CONSTRAINS,		"solvable:constrains"),		/* conda */
+KNOWNID(SOLVABLE_TRACK_FEATURES,	"solvable:track_features"),	/* conda */
 
 KNOWNID(ID_NUM_INTERNAL,		0)
 
--- a/src/policy.c
+++ b/src/policy.c
@@ -846,6 +846,11 @@ pool_buildversioncmp(Pool *pool, Solvable *s1, Solvable *s2)
 static int
 pool_buildflavorcmp(Pool *pool, Solvable *s1, Solvable *s2)
 {
+  unsigned long long tf1 = solvable_lookup_num(s1, SOLVABLE_TRACK_FEATURES, 0);
+  unsigned long long tf2 = solvable_lookup_num(s2, SOLVABLE_TRACK_FEATURES, 0);
+  if (tf1 != tf2)
+    return tf1 > tf2 ? -1 : 1;
+
   const char *f1 = solvable_lookup_str(s1, SOLVABLE_BUILDFLAVOR);
   const char *f2 = solvable_lookup_str(s2, SOLVABLE_BUILDFLAVOR);
   if (!f1 && !f2)


Conan-Kudo:
@wolfv Isn't this basically the same as if `mamba` used the SOLVER_DISFAVOR feature for track_features?

wolfv:
good question -- i don't know . this way it's in the right sort order (from 0 track feature to N) even though I have only seen packages with 0 or 1 track_feature.

mlschroe:
But shouldn't we store the track features instead of just the count? A match_spec can select some track_feature, right?

And isn't conda's minimization about the number of track features after the solving (i.e. some global view), and not about a local track feature count?

wolfv:
As far as I understand, track_features is officially broken, and the way to achieve "variants" is to create mutex packages (which work fine, also with libsolv). track_features are mainly used to de-prioritize the non-default variant (as far as I understand).

mlschroe:
Here's a little example to illustrate what I mean. Given the following packages:
```
Package A:
  depends: B|C
  track_feature: foo,bar

Package B:
  track_feature: foo,bar

Package C:
  track_feature: baz
```
Should `conda install A` install A,B or A,C?

wolfv:
track features really is just an historical artifact. It never worked well for solving, and now is just used to de-prioritize packages ... that's at least how I understand it. Also the manual doesn't specify anything except for track_features weighing down a package, and that "no two packages in a subdir can have the same track_features".

https://github.com/conda-forge/openjdk-feedstock/issues/37#issuecomment-440933865

Maybe @SylvainCorlay can give his opinion?

SylvainCorlay:
Just quoting the original documentation for `track_features` and `features`.

```
========
Features
========

Features are a way to track differences in two packages that have
the same name and version. For example, a feature might indicate
a specialized compiler or runtime, or a fork of a package. The
canonical example of a feature is the mkl feature in Anaconda
Accelerate. Packages that are compiled against MKL, such as
NumPy, have the mkl feature set. The mkl metapackage has the mkl
feature set in track_features, so that installing it installs
the mkl feature. Feature names are independent of package
names---it is a coincidence that mkl is both the name of a
package and the feature that it tracks.

Think of features as belonging to the environment the package is
installed into, not the package itself. When a feature is
installed, conda automatically changes to a package with that
feature if it exists. For example, when the mkl feature is
installed, regular NumPy is removed and the NumPy package with
the mkl feature is installed.  Enabling a feature does not
install any packages that are not already installed, but all
future packages with that feature that are installed into that
environment will be preferred.docs/source/user-guide/tasks/build-packages/features.rst

To install a feature, install a package that tracks it. To remove
a feature, use ``conda remove --features``.

It is a good idea to create a metapackage for track_features.
If you add track_features to a package that also has versions
without that feature, then the versions without that feature will
never be selected, because conda will always add the feature when
it is installed from the track_features specification of your
package with the feature.

EXAMPLE: If you want to create some packages with the feature
debug, you would create several packages with the following
code:

.. code-block:: yaml

   build:
     features:
       - debug

Then you would create a special metapackage:

.. code-block:: yaml

   package:
     # This name doesn't have to be the same as the feature, but can avoid confusion if it is
     name: debug
     # This need not relate to the version of any of the packages with the
     # feature. It is just a version for this metapackage.
     version: 1.0

   build:
     track_features:
       - debug

.. or use ``conda install --features``, blocking on
.. https://github.com/conda/conda/issues/543
```

```
Features
--------

Defines what features a package has. For more information, see
:doc:`features`.

.. code-block:: yaml

   build:
     features:
       - feature1


Track features
--------------

To enable a feature, install a package that tracks that feature.
A package can have a feature, track that feature, or both, or
neither. Usually it is best for the package that tracks a
feature to be a metapackage that does not have the feature. For
more information, see :doc:`features`.

.. code-block:: yaml

   build:
     track_features:
       - feature2
```

SylvainCorlay:
I think that it is fine to merely keep the depriorization behavior since the complete behavior is not documented in the recent versions of conda.

conda's "features/track_features" caused issues with conda's SAT solver.

SylvainCorlay:
Hey, is there any chance of getting this in?

mlschroe:
Fixed with commit c7d13d05a587cec227fa22cc8e4b403e85b778c6

I decided to at least store all the track_features individually instead of the count, so that we can change the way we deal with them later on if there is the need.

I've added a solvable_lookup_count() method for this so that looking up the count of an id array should not be much slower than looking up a num.

I've also extended the parsing of track features so that it also understands an array instead of a string (like the conda parser does).

mlschroe:
And sorry that this took so long, I was somewhat busy with other stuff.

wolfv:
@mlschroe Awesome! Thanks a a lot for this!

Did you have time to take a look at this one as well? https://github.com/openSUSE/libsolv/pull/362

PR #366, opened 2020-02-05
add mips arch support

fangyaling:
add mips arch support


mlschroe:
Those new lines are actually not needed, as the left side matches the right side. That's what libsolv already uses in pool_setarch() if it does not find a matching line.

(This was different in the past, see commit aea953a42d99399938eb6c98a220310aa438c376 which deleted the old mips entries)

PR #368, opened 2020-02-12
Add ENABLE_PYTHON3 to the conditional that triggers building bindings

dralley:


dralley:
See: https://github.com/openSUSE/libsolv/blob/master/bindings/CMakeLists.txt#L11-L13

mlschroe:
No objections, but note that you should only use ENABLE_PYTHON3 if you plan to build *both* the python2 and the python3 bindings.

mlschroe:
(That's also why the ENABLE_PYTHON3 option was hidden, it's not meant to be used if you're building for just one python version.)

dralley:
Thanks for the info, I wasn't aware of that.  Do you think that still makes sense now that Python 2 is EOL?

mlschroe:
I'm fine with adding ENABLE_PYTHON3 to that "if" statement, but I'd prefer to keep ENABLE_PYTHON3 somewhat hidden by not announcing it in an OPTION statement.

mlschroe:
Thanks!

PR #375, opened 2020-04-18
allow win32 to build shared lib

wolfv:


wolfv:
@mlschroe if we could get an release with this and the other changes, that would be really awesome :)

src/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ IF (DISABLE_SHARED)
 ADD_LIBRARY (libsolv STATIC ${libsolv_SRCS})
 ELSE (DISABLE_SHARED)
 ADD_LIBRARY (libsolv SHARED ${libsolv_SRCS})
+TARGET_COMPILE_DEFINITIONS(libsolv PRIVATE SOLV_EXPORTS)
```
mlschroe, line 54: Can we move that to win32/CMakeLists.txt?

wolfv: okaydokey

wolfv: wait, no it's not possible. the trick is that this target has to define `SOLV_EXPORTS` so that `dllexport` is selected, while other targets linking to libsolv have to not define it (so that the macro in `config.h` selects `dllimport`).

So the target needs to be available.

I can make a `if (WIN32)` clause around it or change the name of the compile definition to `WIN32_SOLV_EXPORTS` or `DLL_SOLV_EXPORTS`) if you prefer?

mlschroe: Makes sense. We need the win32/CMakeLists.txt before creating the target.
Ok, so I guess I prefer a `IF (WIN32)` clause in that case.

wolfv: alright!

mlschroe:
libsolv 0.7.12 is released!

wolfv:
awesome! thanks a lot :) we're pushing hard on mamba right now and it's shaping up very nicely. thanks for all your help.

PR #376, opened 2020-04-23
Fix build with rpm headers in non-standard path

mikhirev:
Hello! I added search for `rpm/rpmio.h` to `CMakeLists.txt` to allow building with rpm installed to non-standard prefix.


Conan-Kudo:
Why is this change needed?

mikhirev:
Because building libsolv with rpm installed to non-standard prefix fails with
```
/home/build/libsolv-0.7.10/ext/repo_rpmdb.c:29:10: fatal error: rpm/rpmio.h: No such file or directory
```
because there's no corresponding `-I...` option passed to the compiler.

mikhirev:
Do you need any additional information? Or should I open an issue for this?

mlschroe:
I don't get the commit. In what directory is your `rpm/rpmio.h` file? Why is that directory found with the `FIND_PATH` call?

mikhirev:
In my particular case, rpm is installed in temporary prefix under `$HOME`. That prefix is passed to cmake via `CMAKE_PREFIX_PATH` variable, so `find_path` is able to find headers there.

mlschroe:
Got it, thanks!

PR #379, opened 2020-04-25
dllexport solv_version as well

wolfv:
Unfortunately I didn't see this the first time around. 

I saw that you have a `.ver` file with all exported symbols but quickly googling didn't tell me if we could use this on Windows as well.


wolfv:
ok we need to actually install the `config.h` file otherwise it's not working right.

Would you be OK to move the file to the solv directory as `win32_config.h` and install it on win systems?

wolfv:
@mlschroe are there any objections to this PR?

Conan-Kudo (approved):

PR #384, opened 2020-05-19
conda: take subprio into account for initial sort

wolfv:
@mlschroe what do you think about this?

This produces slightly nicer results for conda environments. Conda has two modes: 

- strict channel priority (always choose package from highest prio channel, works beautifully with libsolv)
- dynamic prio: take highest evr. However, often we have the same versions on multiple channels, and users still expect mamba to choose the one from e.g. conda-forge, so adding a sub-priority and using it in this sort gives a slight preference to those packages from the added channel.

### DIFF
--- a/src/policy.c
+++ b/src/policy.c
@@ -906,10 +906,14 @@ prune_to_best_version(Pool *pool, Queue *plist)
       if (r == 0 && has_package_link(pool, s))
         r = pool_link_evrcmp(pool, best, s);
 #endif
+#ifdef ENABLE_CONDA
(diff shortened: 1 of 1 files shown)


mlschroe:
So the subprio should count stronger than the buildversion/buildflavor? That's a bit odd, it was meant to be used as a tie-break for otherwise identical packages.

Anyway, it is a conda only change, so i don't mind merging this change.

ignatenkobrain:
It depends. Since I'd like to have buildflavor functionality for fedora
repos, it would be nice to have a flag for this.

PR #386, opened 2020-05-25
rpmdb: Also detect rpmdb.sqlite in /usr/share/rpm

cgwalters:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1838691

This is obviously adding to a tower of hacks; what we really
want is for this logic to live in librpm and be standardized.
But in order to fix the immediate issue of detecting the rpmdb
on rpm-ostree based systems (and any others that use `/usr/share/rpm`),
add the sqlite path to detection.

See also http://lists.rpm.org/pipermail/rpm-maint/2017-October/006681.html

### DIFF
--- a/ext/repo_rpmdb_librpm.h
+++ b/ext/repo_rpmdb_librpm.h
@@ -46,7 +46,7 @@ static void
 detect_dbpath(struct rpmdbstate *state)
 {
   state->dbpath = access_rootdir(state, "/var/lib/rpm", W_OK) == -1
-                  && access_rootdir(state, "/usr/share/rpm/Packages", R_OK) == 0
+                  && (access_rootdir(state, "/usr/share/rpm/Packages", R_OK) == 0 || access_rootdir(state, "/usr/share/rpm/rpmdb.sqlite", R_OK) == 0)
                   ? "/usr/share/rpm" : "/var/lib/rpm";
 }
 


ignatenkobrain:
FWIW, the bug is private…

cgwalters:
> FWIW, the bug is private…

Something must have changed in BZ to more aggressively default to private bugs.  That one was marked `fedora_contrib_private` - what the heck?  Just doesn't make any sense at all.

mlschroe:
I wonder if we should just check for the `/usr/share/rpm` directory...

cgwalters:
Is there no way in the libsolv API to override the dbpath?  I thought OpenSUSE switched to `/usr/lib/sysimage/rpm` per http://lists.rpm.org/pipermail/rpm-maint/2017-October/006681.html - but AFAICS there are no patches to do that in https://build.opensuse.org/package/show/zypp:Head/libsolv ?

Conan-Kudo:
@cgwalters They maintain a symlink in `/var/lib/rpm` pointing back to `/usr/lib/sysimage/rpm` because Zypper is hardwired to that path.

cgwalters:
Doing it that way is gross though in `--destdir` style installs, because then we need to have that symlink set up very early just so our own process can find it.  But, I guess we can do it that way.

I think the cleanest fix though per the above BZ is to get the "find the rpmdb location" in rpm itself and have it support the sysimage dir, now that libsolv uses librpm.

Conan-Kudo:
Right, it should just pull the information from `%_dbpath`, since that _is_ set early enough. The only issue is when the `%_dbpath` differs between host and guest/installroot.

PR #391, opened 2020-06-02
Fix all memory errors found in 'make test'

### LINKED ISSUE #389: Use-after-free with solv_xfopen_buf
In various test cases there is a buffer being used after it goes out of scope.
`testcase.c:2070`:
```
	      if (!strcmp(pieces[4], "<inline>"))
		{
		  char *idata = read_inline_file(fp, &buf, &bufp, &bufl);
		  rdata = "<inline>";
		  rfp = solv_xfopen_buf(rdata, &idata, 0, "rf");
		}
```
But then `solv_xfopen_buf` in `solv_xfopen.c:848` takes `idata` as `bufp`:
```
  bc->bufp = bufp;
```
So `bc` (the bufcookie) associated with `rfp` keeps pointing to `idata` (the `char *` itself; not the data that has been allocated) after the block ends (via `cookieopen`). Later in `testcase.c:2087` again:
```
		  testcase_add_testtags(repo, rfp, 0);
```
Tries to read the old `idata` pointer (via `rfp` AKA `fp`) `repo_testcase.c:557`:
```
      if (!fgets(linep, aline - (linep - line), fp))
```

This is 1 of 20 failures of the 23 tests when linked against the address sanitizer `libasan`; I assume the other 19 are very similar. This one in particular was the `yumobs` test.

To replicate the failure:
```
$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 767aa665..b16bdb1c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,18 @@ ENDIF (NOT PKGCONFIG_INSTALL_DIR)
 # CONFIGURATION                                                    #
 ####################################################################
 
+
+# AddressSanitize
+set(CMAKE_C_FLAGS_ASAN
+    "-fsanitize=address -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer -g -O1"
+    CACHE STRING "Flags used by the C compiler during AddressSanitizer builds."
+    FORCE)
+set(CMAKE_CXX_FLAGS_ASAN
+    "-fsanitize=address -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer -g -O1"
+    CACHE STRING "Flags used by the C++ compiler during AddressSanitizer builds."
+    FORCE)
+
+
 # where to look first for cmake modules, before ${CMAKE_ROOT}/Modules/ is checked
 SET (CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
 INSTALL( FILES ${CMAKE_MODULE_PATH}/FindLibSolv.cmake DESTINATION ${CMAKE_INSTALL_PREFIX}/share/cmake/Modules )
```
and then
```
cmake -DCMAKE_BUILD_TYPE=ASAN ..
make test
```

All line numbers from 990f3392.

Raw output (decoded above):
```
$ cd test/testcases/yumobs
$ ../../..//build/tools/testsolv split.t 
=================================================================
==13317==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7ffe0a46e200 at pc 0x7f2bbcf16fb0 bp 0x7ffe0a46d570 sp 0x7ffe0a46d560
READ of size 8 at 0x7ffe0a46e200 thread T0
    #0 0x7f2bbcf16faf in cookie_bufread /.../libsolv/ext/solv_xfopen.c:809
    #1 0x7f2bbcc3cb51 in _IO_new_file_underflow (/lib64/libc.so.6+0x83b51)
    #2 0x7f2bbcc3ddd5 in __GI__IO_default_uflow (/lib64/libc.so.6+0x84dd5)
    #3 0x7f2bbcc3099b in __GI__IO_getline_info (/lib64/libc.so.6+0x7799b)
    #4 0x7f2bbcc2f8e9 in _IO_fgets (/lib64/libc.so.6+0x768e9)
    #5 0x7f2bbcf7b9fb  (/lib64/libasan.so.6+0x409fb)
    #6 0x7f2bbcf297e0 in testcase_add_testtags /.../libsolv/ext/repo_testcase.c:557
    #7 0x7f2bbcf237f9 in testcase_read /.../libsolv/ext/testcase.c:2087
    #8 0x403086 in main /.../libsolv/tools/testsolv.c:149
    #9 0x7f2bbcbe0041 in __libc_start_main (/lib64/libc.so.6+0x27041)
    #10 0x40240d in _start (/.../libsolv/build/tools/testsolv+0x40240d)

Address 0x7ffe0a46e200 is located in stack of thread T0 at offset 192 in frame
    #0 0x7f2bbcf2297a in testcase_read /.../libsolv/ext/testcase.c:1965

  This frame has 10 object(s):
    [48, 52) 'bufl' (line 1968)
    [64, 68) 'prio' (line 2051)
    [80, 84) 'subprio' (line 2051)
    [96, 100) 'what' (line 2163)
    [112, 116) 'what' (line 2297)
    [128, 136) 'buf' (line 1967)
    [160, 168) 'bufp' (line 1967)
    [192, 200) 'idata' (line 2072) <== Memory access at offset 192 is inside this variable
    [224, 256) 'autoinstq' (line 1979)
    [288, 320) 'q' (line 2216)
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-use-after-scope /.../libsolv/ext/solv_xfopen.c:809 in cookie_bufread
Shadow bytes around the buggy address:
  0x100041485bf0: 00 00 00 00 00 00 f1 f1 f1 f1 f1 f1 00 00 00 00
  0x100041485c00: 00 f3 f3 f3 f3 f3 00 00 00 00 00 00 00 00 00 00
  0x100041485c10: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100041485c20: 00 00 00 00 00 00 00 00 f1 f1 f1 f1 f1 f1 04 f2
  0x100041485c30: 04 f2 04 f2 04 f2 04 f2 00 f2 f2 f2 00 f2 f2 f2
=>0x100041485c40:[f8]f2 f2 f2 00 00 00 00 f2 f2 f2 f2 00 00 00 00
  0x100041485c50: f3 f3 f3 f3 00 00 00 00 00 00 00 00 00 00 00 00
  0x100041485c60: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100041485c70: f1 f1 f1 f1 f1 f1 04 f2 00 f2 f2 f2 00 00 f2 f2
  0x100041485c80: 00 00 00 00 f2 f2 f2 f2 00 00 00 00 f2 f2 f2 f2
  0x100041485c90: 00 00 00 00 f2 f2 f2 f2 00 00 00 00 f3 f3 f3 f3
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==13317==ABORTING
```

AaronDMarasco:
Fixes #389 as well as a handful of memory leaks

### DIFF
--- a/ext/solv_xfopen.c
+++ b/ext/solv_xfopen.c
@@ -796,6 +796,7 @@ solv_xfopen_iscompressed(const char *fn)
 struct bufcookie {
   char **bufp;
   size_t *buflp;
+  char *curmem;
   char *freemem;
   size_t bufl_int;
 };
@@ -860,7 +861,11 @@ solv_xfopen_buf(const char *fn, char **bufp, size_t *buflp, const char *mode)
     }
   fp = cookieopen(bc, mode, cookie_bufread, cookie_bufwrite, cookie_bufclose);
   if (!strcmp(mode, "rf"))	/* auto-free */
-    bc->freemem = *bufp;
+    { /* need to capture *bufp locally #389 */
+      bc->bufp = &bc->curmem;
+      bc->curmem = *bufp;
+      bc->freemem = *bufp;
+    }
   if (!fp)
     {
       if (*mode == 'w')
--- a/src/pool.c
+++ b/src/pool.c
@@ -62,6 +62,7 @@ pool_create(void)
   queue_init(&pool->vendormap);
   queue_init(&pool->pooljobs);
   queue_init(&pool->lazywhatprovidesq);
+  pool->considered = 0;
 
 #if defined(DEBIAN)
   pool->disttype = DISTTYPE_DEB;
@@ -117,6 +118,11 @@ pool_free(Pool *pool)
   queue_free(&pool->vendormap);
   queue_free(&pool->pooljobs);
   queue_free(&pool->lazywhatprovidesq);
+  if (pool->considered)
+    {
+      map_free(pool->considered);
+      solv_free(pool->considered);
+    }
   for (i = 0; i < POOL_TMPSPACEBUF; i++)
     solv_free(pool->tmpspace.buf[i]);
   for (i = 0; i < pool->nlanguages; i++)
--- a/src/selection.c
+++ b/src/selection.c
@@ -163,6 +163,7 @@ selection_flatten(Pool *pool, Queue *selection)
       selection->elements[0] = SOLVER_SOLVABLE | SOLVER_NOAUTOSET;
       selection->elements[1] = q.elements[0];
     }
+    queue_free(&q);
 }
 
 /* only supports simple rels plus REL_ARCH */
--- a/tools/testsolv.c
+++ b/tools/testsolv.c
@@ -150,6 +150,7 @@ main(int argc, char **argv)
 	  if (!solv)
 	    {
 	      pool_free(pool);
+	      queue_free(&job);
 	      exit(resultflags == 77 ? 77 : 1);
 	    }
 	  if (reusesolv)


ext/solv_xfopen.c
```diff
@@ -796,6 +796,7 @@ solv_xfopen_iscompressed(const char *fn)
 struct bufcookie {
   char **bufp;
   size_t *buflp;
+  char *curmem;
```
AaronDMarasco, line 799: A new variable was needed because reads will move the pointer and then the `solv_free(freemem)` elsewhere will point _into_ the buffer, not the base address of the buffer.

src/pool.c
```diff
@@ -117,6 +118,11 @@ pool_free(Pool *pool)
   queue_free(&pool->vendormap);
   queue_free(&pool->pooljobs);
   queue_free(&pool->lazywhatprovidesq);
+  if (pool->considered)
+    {
+      map_free(pool->considered);
+      solv_free(pool->considered);
+    }
```
mlschroe, line 125: We can't do that. We did not allocate pool->considered, so we do not know how to free it.

mlschroe:
Cherry-picked.

mlschroe:
(And yes, I do think the pool->considered interface sucks, but it's way too late to change it as it is already used in a couple of projects like perl-BSSolv or Fedora's libdnf.)

src/pool.c
AaronDMarasco: If `pool.h` defines it as `Map *considered`, are you saying some implementations are using it as a free scratch space and not really a `Map *`?

Either way, I think the initialization fix in `pool_create()` should probably still be included.

mlschroe: No, but the Map it points to can be part of some bigger struct, so we can't do a solv_free() on the pointer. E.g.:
```
struct my_pool {
  Pool *pool;
  Map considered;
};
```

There actually *is* a calloc in pool_create():
```
  pool = (Pool *)solv_calloc(1, sizeof(*pool));
```

PR #392, opened 2020-07-01
small conda startswith improvement

wolfv:
We had an issue report where

`ucx-py=0.15*`

did not match a package version like this

`ucx-py 0.15.0a200630+g13ff111`

This is my attempt to fix it. Do you see any problems with this approach @mlschroe?


mlschroe:
Looks good, thanks!

For reference: conda does this in the startswith() function in conda/models/version.py

jakirkham:
Thank you 😀

PR #398, opened 2020-08-21
Add extra arg to write_testcase for libzypp

bzeller:
In order for libzypp/zypper to write its own testcase informations we need
a way to get the repo data filenames. This patch adds a extra argument
to receive the filenames for each repository.

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -3685,7 +3685,7 @@ rb_eval_string(
   }
 
   bool write_testcase(const char *dir) {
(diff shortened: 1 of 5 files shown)


mlschroe:
I'd prefer to put the extra information you need in the repo meta area (SOLVID_META). Would that also work for you?

bzeller:
> I'd prefer to put the extra information you need in the repo meta area (SOLVID_META). Would that also work for you?

So it would be part of the testcase.t file where all the repos are listed?
(As long as I can query the information from the resulting testcase, or easily read it it works for me )

mlschroe:
No longer needed.

PR #401, opened 2020-09-22
Updateinfo metadata with multiple collections in a single pkglist (nevra/nsvca construction)

kontura:
In rhel there are updateinfo metadata with multiple collections in a single pkglist, example:
```
<pkglist>
  <collection>
    <module name="mod1" stream="master" version="2" context="c1" arch="x86_64"/>
    <package name="pkg-a" version="1" release="2" epoch="0" arch="x86_64" src="http://www.foo.org">
      <filename>pkg-a-1-2.spec</filename>
    </package>
  </collection>
  <collection>
    <module name="mod2" stream="master" version="2" context="c2" arch="x86_64"/>
    <package name="pkg-b" version="1" release="2" epoch="0" arch="x86_64" src="http://www.foo.org">
      <filename>pkg-b-1-2.spec</filename>
    </package>
    <package name="pkg-lib" version="1" release="2" epoch="0" arch="x86_64" src="http://www.foo.org">
      <filename>pkg-lib-1-2.spec</filename>
    </package>
  </collection>
</pkglist>
```

Currently the information about collections is lost during parsing, all packages end up together and all modules as well, I am trying to fix that by adding the `collectionlist` which contains modules and packages for each collection.
```
update:collectionlist:
  update:module:
    mod1:master:2:c1:x86_64
  update:collection:
    pkg-a-1-2.x86_64

  update:module:
    mod2:master:2:c2:x86_64
  update:collection:
    pkg-b-1-2.x86_64
    pkg-lib-1-2.x86_64
```

I don't really like the nevra and nsvca construction, the approach in conflicting PR https://github.com/openSUSE/libsolv/pull/402 is much cleaner but I am not sure if correct. In case both are wrong can you point me in the right direction?

### DIFF
--- a/ext/repo_updateinfoxml.c
+++ b/ext/repo_updateinfoxml.c
@@ -110,7 +110,7 @@ struct parsedata {
   Id handle;
   Solvable *solvable;
   time_t buildtime;
-  Id collhandle;
+  Id pkghandle;
   struct solv_xmlparser xmlp;
   struct joindata jd;
 };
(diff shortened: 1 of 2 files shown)


kontura:
Closing in favor of https://github.com/openSUSE/libsolv/pull/402.

PR #402, opened 2020-09-22
Updateinfo metadata with multiple collections in a single pkglist (using only ids)

kontura:
This is a conflicting PR for https://github.com/openSUSE/libsolv/pull/401 that tries to fix the same problem with updateinfo metadata with multiple collections in a single pkglist.

I would prefer this approach but I don't know if it is allowed/correct.
I am able to get the information about collections in libdnf using two nested `dataiterators` but when I try to `dumpsolv`  such a solv file it doesn't work, reporting:
```
data_read_id_max: id too large (14/7)
could not read repository: data_read_id_max: id too large (14/7)
```
I think it has something to do with the `schematas` but I can't seem to figure it out.

### DIFF
--- a/ext/repo_updateinfoxml.c
+++ b/ext/repo_updateinfoxml.c
@@ -110,7 +110,7 @@ struct parsedata {
   Id handle;
   Solvable *solvable;
   time_t buildtime;
-  Id collhandle;
+  Id pkghandle;
   struct solv_xmlparser xmlp;
   struct joindata jd;
 };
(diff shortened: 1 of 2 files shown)


j-mracek:
Hi Michael @mlschroe, please could you help us with a support of multiple contexts in libsolv. It is very critical for us (errata applicability). @kontura prepare two PRs that resolves the same problem. This one look as a better option to me, but it not working and we don't know why. Probably you have a some idea what's wrong. As I mentioned, the problem is critical to us therefore @kontura prepared an alternative that is functional: https://github.com/openSUSE/libsolv/pull/401. Thank you very much.

mlschroe:
Eep, sorry. Forgot about that one. Will work on it ASAP.

mlschroe:
Could you please point me to such an updateinfo,xml file with multiple collection entries? Thanks!

mlschroe:
Found the problem. Working on a fix...

kontura:
> Found the problem. Working on a fix...

That is great. Thank you.

Just in case here is a full updateinfo.xml file from j-mracek's repo with multiple collections: https://github.com/j-mracek/module_repos/blob/master/advisories/multicontext/updateinfo.xml

mlschroe:
Turned out this was a problem with the handling of nested flexarrays in the solv file writer.

Ok if I merge this pull request now?

kontura:
Yes, it is working now. :slightly_smiling_face:

PR #403, opened 2020-10-08
package: Set LD_LIBRARY_PATH for tests to work against shared library

Conan-Kudo:
When a shared library is built, we need to have the test suite attempt to
use our newly built shared library. Without this, it fails with unresolved
symbols.

### DIFF
--- a/package/libsolv.spec.in
+++ b/package/libsolv.spec.in
@@ -263,6 +263,11 @@ export NO_BRP_STRIP_DEBUG=true
 %endif
 
 %check
+%if %{with shared}
+# The test suite doesn't automatically know to look at the "built"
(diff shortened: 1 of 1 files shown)


mlschroe:
Actually the test suite *does* know how to use the newly built library, that's a cmake feature. So I'm kind of puzzled how you get errors.

Conan-Kudo:
@mlschroe It did not work until I made that change for my build here: https://build.opensuse.org/package/rdiff/home:Pharaoh_Atem:DNF_SUSE:DNF5/libsolv?linkrev=base&rev=2

Conan-Kudo:
@mlschroe Note this is only broken on openSUSE, as I don't have this problem on Fedora or Mageia.

PR #407, opened 2020-11-01
Home rpmdb

fepitre:
rpm on Debian is patched to not use the default system-wide rpm database. Instead, the dbpath is hardcoded to ~/.rpmdb. This is based on previous original work of author in https://lists.debian.org/debian-devel/2019/09/msg00218.html.

### DIFF
--- a/ext/libsolvext.ver
+++ b/ext/libsolvext.ver
@@ -54,6 +54,7 @@ SOLV_1.0 {
 		rpm_query_num;
 		rpm_stat_database;
 		rpm_state_create;
+		rpm_state_create_real;
 		rpm_state_free;
 		solv_fmemopen;
 		solv_verify_sig;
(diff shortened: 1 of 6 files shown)


Conan-Kudo (changes requested):
NAK. The "homedir rpmdb" functionality was rejected by RPM years ago and still doesn't exist today, so this stuff built on it is not acceptable.

fepitre:
OK so I guess there is no way for having it here even if `RPM` project itself does not want such setup?

Conan-Kudo:
Yeah, the RPM project advised the Debian folks to drop those changes entirely because it causes unexpected breakages.

fepitre:
I agree with you that Debian should use default setup and I was not aware of previous attempt. Just in case, if you would have the thread from RPM as reference, I could try to discuss with Debian folks too.

Conan-Kudo:
Unfortunately, I don't have that thread because I think it was back when RPM used its own Trac instance, which has been gone for years...

fepitre:
Sure no problem. Thank you for your feedback.

PR #409, opened 2020-11-13
Parsing comps - add default, langonly and parse boolean values

pkratoch:


ext/repo_comps.h
```diff
@@ -5,4 +5,9 @@
  * for further information
  */
 
+#include <stdbool.h>
+
 extern int repo_add_comps(Repo *repo, FILE *fp, int flags);
+
+const bool COMPS_DEFAULT_USERVISIBLE = true;
+const bool COMPS_DEFAULT_DEFAULT = false;
```
Conan-Kudo, line 13: Is this something that should be exposed in the API or should be internal to the code? If the latter, shouldn't this be in the `repo_comps.c` file?

Conan-Kudo (changes requested):

ext/repo_comps.h
pkratoch: It doesn't need to be exposed, so I moved it to ``repo_comps.c``.

Conan-Kudo (approved):

Conan-Kudo:
@mlschroe Could you please review this? It's needed for work in libdnf.

mlschroe:
What happened to libcomps?

pkratoch:
> What happened to libcomps?

Libcomps stays available as a tool for managing groups, but we want to use libsolv in dnf, so that we will possibly have one less dependency.

PR #434, opened 2021-03-01
Add flag to skip conda v2 packages when parsing conda repos

tl-hbk:
conda has a config setting `use_only_tar_bz2` that allows you to skip all packages that don't end with `tar.bz2` that `mamba` doesn't currently support.

https://github.com/mamba-org/mamba/issues/666

This PR adds a flag for calling `repo_add_conda` to skip conda v2 type packages when parsing a repo

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -258,7 +258,7 @@ parse_packages2(struct parsedata *pd, struct solv_jsonparser *jp)
 }
 
 static int
-parse_main(struct parsedata *pd, struct solv_jsonparser *jp)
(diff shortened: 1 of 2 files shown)


mlschroe (changes requested):
Could you please change the flag to use `(1<<8)`? The first 8 bits are reserved for generic bits, i.e. REPO_USE_ROOTDIR, REPO_NO_LOCATION and the like.

tl-hbk:
@mlschroe done

mlschroe:
Thanks!

Issue #435, opened 2021-03-04
Issue with regex.h polyfill for windows build


tl-hbk:
I noticed that mamba which uses libsolv was returning unexpected results on windows but working fine on linux and after some digging I think the issue is that the regex polyfill doesn't return the same results.  More details are in the mamba issue [here](https://github.com/mamba-org/mamba/issues/725#issuecomment-790836071).  

This is the code snippet I used to confirm it.  On windows `regexmatch` which I copied from `conda.c`, has been returning `1` for any combination of strings I pass it.

```
#include <stdio.h>
#include <stdlib.h>
#include <stdarg.h>
#include <unistd.h>
#include <string.h>
#include <sys/types.h>
#include <regex.h>

#include "util.h"

static int
regexmatch(const char *evr, const char *version, size_t versionlen, int icase)
{
  regex_t reg;
  char *buf = solv_malloc(versionlen + 1);
  int r;

  memcpy(buf, version, versionlen);
  buf[versionlen] = 0;
  if (regcomp(&reg, buf, REG_EXTENDED | REG_NOSUB | (icase ? REG_ICASE : 0)))
    {
      solv_free(buf);
      return 0;
    }
  r = regexec(&reg, evr, 0, NULL, 0);
  regfree(&reg);
  solv_free(buf);
  return r == 0;
}

int
main(int argc, char **argv)
{
    char* evr = "2020.03.04";
    char* version = ".*\\.dev.*";
    int res = regexmatch(evr, version, strlen(version), 0);
    fprintf(stderr, "regexmatch: %d\n", res);
    exit(0);
}
```

tl-hbk:
Like I said in the linked mamba issue I'm not too familiar with C so I'm not sure how difficult this is to fix but having looked at musl docs it doesn't seem like there's [official windows support](https://wiki.musl-libc.org/supported-platforms.html) listed.  I did find PCRE which seems to have implement the [POSIX regex interface](https://www.pcre.org/original/doc/html/pcreposix.html), already have [conda packages](https://anaconda.org/conda-forge/pcre/files) which is how the windows builds are being built from what I can tell, and is [BSD licensed](https://www.pcre.org/licence.txt) @wolfv

PR #436, opened 2021-03-05
Fix regex bug on windows

tl-hbk:
Fixes this issue https://github.com/openSUSE/libsolv/issues/435 

Details in this thread https://github.com/mamba-org/mamba/issues/725#issuecomment-791353298

### DIFF
--- a/win32/regex.h
+++ b/win32/regex.h
@@ -12,7 +12,8 @@ extern "C" {
 #define CHARCLASS_NAME_MAX 14
 #define RE_DUP_MAX 255
 
-typedef size_t regoff_t;
+#include <BaseTsd.h>
+typedef SSIZE_T regoff_t;
 // #include <bits/alltypes.h>
 
 typedef struct re_pattern_buffer {


mlschroe:
What"s wrong with ssize_t?

tl-hbk:
@mlschroe it would fail to build when I tried to use `ssize_t` with msvc

mlschroe:
Even if you do a `#include <sys/types.h>`?

tl-hbk:
Yes

tl-hbk:
@mlschroe would you prefer `SSIZE_T` be used in this case?

mlschroe:
Yes. I think `long` is just 32bit and thus wrong for _WIN64. (I'm not sure about this, that's just what searching around the internet came up with.)

tl-hbk:
@mlschroe change made

mlschroe:
Thanks!

tl-hbk:
With your comment mentioning `long` being 32bit for windows, I noticed that `long` is used here 
https://github.com/openSUSE/libsolv/blob/master/win32/tre.h#L79-L82 and all the calls to `ALIGN` as well
https://github.com/openSUSE/libsolv/blob/master/win32/regexec.c#L245-L254

I'm assuming this needs to be fixed?

PR #438, opened 2021-03-08
Update long usage in regex polyfill

tl-hbk:
It was brought up in this PR https://github.com/openSUSE/libsolv/pull/436 that `long` is just 32 bit on windows and so I thought this might be necessary.  Please let me know if that's not the case.

EDIT: I also see `long` used here too https://github.com/openSUSE/libsolv/blob/master/win32/regcomp.c#L102-L103 but am more hesitant to change that as I'm not really sure why it's `long` here when `code_min` and `code_max` are defined as `int` here https://github.com/openSUSE/libsolv/blob/master/win32/regcomp.c#L49-L50 and all the usage in `regcomp.c` seem to treat them as `int`

### DIFF
--- a/win32/regexec.c
+++ b/win32/regexec.c
@@ -231,7 +231,7 @@ tre_tnfa_run_parallel(const tre_tnfa_t *tnfa, const void *string,
     pbytes = sizeof(*reach_pos) * tnfa->num_states;
     xbytes = sizeof(regoff_t) * num_tags;
     total_bytes =
-      (sizeof(long) - 1) * 4 /* for alignment paddings */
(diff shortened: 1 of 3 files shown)


mlschroe:
Could you please change all the int64_t to size_t so that they stay 32bit on 32bit windows?
(Maybe also uintptr_t, but that's up to you)

mlschroe:
about code_min/max: yes, I also think that can stay as 'long', as TRE_CHAR_MAX is 0x10ffff.

tl-hbk:
Change made can you take a look again?

mlschroe:
Thanks!

PR #439, opened 2021-03-11
Fix memory leak of solv->choicerules_info

kontura:
`solv->choicerules_info` are always allocated at the beginning of
`solver_addchoicerules`, free them before overriding.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -3339,7 +3339,6 @@ solver_addchoicerules(Solver *solv)
       return;
     }
   now = solv_timems(0);
-  solv->choicerules_info = solv_calloc(solv->pkgrules_end, sizeof(Id));
   queue_init(&q);
   queue_init(&qi);
   queue_init(&qcheck);


Conan-Kudo (approved):

mlschroe:
Fine with me, but `solv->choicerules_info` is already free'd at the beginning of solver_solve() and solver_addchoicerules() should only get called once. So I'm wondering if this is a leak you found with some tool like valgrind and I just don't see how it can happen?

kontura:
I did find it using valgrind, but doesn't it happen every time (or the one time) `solver_addchoicerules()` is called? Since just in the scope of the function it is first allocated and then overwritten or am I missing something?

mlschroe:
Ah, now I see what you mean. The first allocation is simply bogus and should be removed. This was forgotten in commit 178203feb32c796b3b0c18057fc69d0b4a96c53c, the old code worked by first allocating an oversized area and later shrinking the info block to the needed size.

Do you want to update the pull request or should I delete the line myself?

kontura:
Alright, I have updated the PR to remove the line.

mlschroe:
Thanks!

PR #442, opened 2021-03-29
Fix memory leaks

jrohel:
The PR fixes two memory leaks.

Note:
Be honest. The function `pool_deb_get_autoinstalled` seems strange to mee. Is the `while (bufl - l < 1024)` loop dead code?


mlschroe:
I think the while loop was kinda copied from repo_add_debpackages(). The origin of the while loop seems to be commit 30b06921106ba2202da36cb1d45e08c95b4daf43

PR #443, opened 2021-03-29
Fix segfault in resolve_jobrules

tl-hbk:
In the for loop in `resolve_jobrules`, `selectandinstall` is called and sometimes rules are added and a `realloc` is required if there's no more memory available in the allocated block.  `r--` would just decrement the old pointer but the `realloc` could return a completely different block of memory.

Issue this was found in for reference https://github.com/mamba-org/boa/issues/146#issuecomment-809374089


mlschroe:
Good catch! I think I'll change this to simply do a
```
  r = solv->rules + i;    /* selectandinstall may have added more rules */
```
after the selectandinstall() call in the code.

mlschroe:
Awesome, thanks!

PR #444, opened 2021-03-30
Fix: lzopen, zstdopen, zchunkopen

jrohel:
- Fix: lzopen and zstdopen: don't close supplied fd
  In the case of a `calloc` memory allocation failure or failure of any of the other functions, the `fp` file stream has been closed.  And the stream closed the associated supplied `fd` file descriptor.
   The caller then closed the closed descriptor again (eg `curlfopen` in `repoinfo_download.c`: `if (!Fp) close (fd);`) -> double_close. In multi threaded software risk of race conditions. Another thread open another file and reuse just closed file descriptor. The caller then closes the file of another thread.

- Fix: zchunkopen: resources leaks, don't close supplied fd
  System variant:
    - resource leaks when `zck_init_read` or `zck_init_write` fails
    - supplied fd will be closed if `zck_create` fails
    
  Libsolv limited zchunk implementation:
   - resource leak when `strcmp(mode, "r") != 0`
   - supplied fd will be closed if `solv_zchunk_open` fails (Fix is thread unsafe. However the original version caused a double-close in the caller function and was thread unsafe too.)

- Fix: lzopen, zstdopen, zchunkopen: Inconsistent file descriptor tests
  Example: The caller passes a valid `path` and an invalid `fd = -2`. It is not caught by the `(!path && fd < 0)` check and the invalid `fd` was used for `fdopen` (because fd != -1) instead of using `path`.

Note:
There are memory allocation inconsistencies in the code. For example, `lzopen` uses` calloc` and `zstdopen` uses` solv_calloc`.
On error: `calloc` returns NULL while` solv_calloc` calls `solv_oom` (` abort`, `exit`). Thus, the behavior of `solv_xfopen` and` solv_xfopen_fd` is different for different types of compression.
In my opinion, the code should be unified. (How does `bzopen` behave in low memory conditions?)

### DIFF
--- a/ext/solv_xfopen.c
+++ b/ext/solv_xfopen.c
@@ -199,7 +199,7 @@ static LZFILE *lzopen(const char *path, const char *mode, int fd, int isxz)
   if (ret != LZMA_OK)
     {
       fclose(fp);
-      free(lzfile);
+      solv_free(lzfile);
       return 0;
     }
   return lzfile;
@@ -232,7 +232,7 @@ static int lzclose(void *cookie)
     }
   lzma_end(&lzfile->strm);
   rc = fclose(lzfile->file);
-  free(lzfile);
+  solv_free(lzfile);
   return rc;
 }
 
@@ -432,7 +432,7 @@ static int zstdclose(void *cookie)
       ZSTD_freeDStream(zstdfile->dstream);
     }
   rc = fclose(zstdfile->file);
-  free(zstdfile);
+  solv_free(zstdfile);
   return rc;
 }
 
--- a/ext/solv_xfopen.c
+++ b/ext/solv_xfopen.c
@@ -165,7 +165,7 @@ static LZFILE *lzopen(const char *path, const char *mode, int fd, int isxz)
   LZFILE *lzfile;
   lzma_ret ret;
 
-  if (!path && fd < 0)
+  if ((!path && fd < 0) || (path && fd >= 0))
     return 0;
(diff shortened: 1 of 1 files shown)


mlschroe:
I don't think the "inconsistent file descriptor tests" commit is needed. Those are static functions, it is not possible to pass a valid path with fd set to -2.

jrohel:
@mlschroe I prefer consistent code and problems prevention.
But you're right. In this case, the functions cannot be called from outside.
So, if you want, I will remove the "inconsistent file descriptor tests"  commit. Do you want?

mlschroe:
How about this: we change the guard at the start of the functions from
```
  if (!path && fd < 0)
    return 0;
```
to
```
  if ((!path && fd < 0) || (path && fd >= 0))
    return 0;
```
for extra problem prevention. Then we do not need an extra `usefd` variable but can check if path is 0 or not. I.e. the checks would not be `if (fd != -1)` but `if (!path)`.

mlschroe:
And the calloc() in lzopen should be solv_calloc(), of course. I copied the code from another project and forgot to change the call. (And solv_calloc() will never return NULL, of course.)

mlschroe:
(I already fixed the calloc call)

jrohel:
- rebased
- Inserted commit `Fix: Use "solv_free" for memory allocated by "solv_calloc"`.
- The guard was changed as you suggested.

mlschroe:
Thanks!

PR #445, opened 2021-04-08
Fix: Memory leaks in SWIG generated code (for Python)

jrohel:
There were memory leaks in the `Chksum_from_bin`, `Chksum_add`, `SolvFp_write` functions wrapper for Python

The problem was in freearg typemap argument defined in "solv.i". Therefore, the typemap was not applied.

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -63,7 +63,7 @@ typedef struct {
   $2 = size;
 }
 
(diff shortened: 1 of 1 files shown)


jrohel:
LGTM

mlschroe:
Indeed. The argument was changed from int to size_t in commit 7d11163abb0a06b717d497526b5305677dad6cbd, but the freearg typemap was not modified. Thanks!

PR #449, opened 2021-06-02
enable build timestamp maximization

wolfv:


Conan-Kudo (changes requested):
Note that as this stands, this logic would affect _all_ repo types using libsolv if CONDA was enabled. I'm not sure that's the intent, though I'm honestly not sure this is a bad behavior from a generic perspective, and it might make sense to just unconditionally do it.

mlschroe:
Yes, the comparison should only be done if the disttype is conda.

PR #454, opened 2021-06-23
Add mechanisms to ignore pkgs for weak dependencies

j-mracek:
It adds mechanism to avoid installation of certain packages
to satisfy recommends and supplemented.


j-mracek:
The patch will be primary use in libdnf/dnf for user configuration weak_excludes, where a user specify name of packages (optionally with EVR, arch, including globs), to prevent packages be installed as a weak dependency (Recommends, Supplements).

Also I try to create possibility to automatically detect unmet weak dependency. The mechanism uses `SOLVER_WEAK_EXCLUDE` but in future there can be another implementation.

doc/libsolv-bindings.txt
```diff
@@ -2076,6 +2076,10 @@ Avoid the specified packages if the solver encounters an alternative. This
 can also be used to block recommended or supplemented packages from being
 installed.
 
+*SOLVER_WEAK_EXCLUDE*::
+Avoid the specified packages to satisfy recommended or supplemented dependencies.
+Unlike SOLVER_DISFAVOR, it did not interfer with other rules.
```
dmach, line 2081: replace `did not interfer` with `does not interfere`

src/solver.h
```diff
@@ -208,6 +208,8 @@ struct s_Solver {
   Map allowuninstallmap;		/* ok to uninstall those */
   int allowuninstall_all;
 
+  Map weakexcludemap;		/* Do not use them for supplements and recommends */
```
dmach, line 211: The comment doesn't make sense to me.
Don't we want to actually use weak excludes on Supplements and Recommends?

dmach (changes requested):

src/solver.h
Conan-Kudo: I would expect that we'd want to, yes...

mlschroe:
I don't think "weak excludes" is a good name, because your weak excludes is different from "excludes, but may be broken" (implemented by SOLVER_DISFAVOR).

I'd prefer some different name like SOLVER_EXCLUDEFROMWEAK. Suggestions welcome.

Conan-Kudo:
@j-mracek Can you rebase this so that the upstream commits aren't included in this PR? It makes merging this not possible.

j-mracek:
@mlschroe May I ask you for your opinion about the pull request? It was enhanced according your suggestions and from community. Is there any chance to get it into libsolv or do you think about an alternative implementation?

j-mracek:
@mlschroe No problem and thanks for your comments

j-mracek:
@mlschroe Do you have a plan for the next release of libsolv? I will be very thankful to see the new version of libsolv with the PR.

PR #459, opened 2021-07-06
Implement strict repo priority

### LINKED ISSUE #458: How to implement 'strict channel priority'?
Conda has a "strict channel priority" where it will only take into account a package from a lower-priority channel if the package doesn't appear in _any_ higher prio channel.

We're currently assigning a repo priority and sub-priority in mamba. However, it's not enough because the repo priority is only checked on the (already) filtered Queue if I understand correctly. 

So for example, suppose we have

channel a:
- numpy 1.0

channel b:
- numpy 0.1

And b has a lower priority than a, I could still install `numpy 0.1` by asking for it specifically `mamba install numpy==0.1` for example. Conda would throw a solver error here.

I could go through all solvables, find those from the highest-prio repo and add them to a ` SOLVER_SOLVABLE_ONE_OF` job but that doesn't strike me as the cleanest solution.

Any ideas how to filter the lower-priority ones out easily?

adriendelsalle:
Description
---

Implement strict repo priority
Add specific rule
Add solver flag
Add problem message/debug

Closes #458

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -4165,24 +4165,21 @@ void solver_addstrictrepopriorules(struct s_Solver *solv, Map *addedmap)
 		{
 		Solvable *s2 = pool->solvables + p2;
 		if (s->name != s2->name)
-		{
-		MAPCLR(&priomap, p2);
 			continue;
-		}
 		if (s2->repo->priority > max_prio)
 			max_prio = s2->repo->priority;
 		}
 		
-    FOR_PROVIDES(p2, pp2, s->name)
+    	FOR_PROVIDES(p2, pp2, s->name)
 		{
 		Solvable *s2 = pool->solvables + p2;
-		if (!MAPTST(&priomap, p2))
-	    continue;
-		MAPCLR(&priomap, p2);
+        if (s->name != s2->name || !MAPTST(&priomap, p2))
+          continue;
+        MAPCLR(&priomap, p2);
 		if (pool->installed && s2->repo == pool->installed)
 			continue;
-		if (s2->repo->priority < max_prio)
-			solver_addrule(solv, -p2, 0, 0);
+        if (s2->repo->priority < max_prio)
+          solver_addrule(solv, -p2, 0, 0);
 		}
 	}
 	solv->strictrepopriorules_end = solv->nrules;
--- a/src/rules.c
+++ b/src/rules.c
@@ -2992,6 +2992,12 @@ solver_ruleinfo(Solver *solv, Id rid, Id *fromp, Id *top, Id *depp)
 	*fromp = -r->p;
       return SOLVER_RULE_BLACK;
     }
+  if (rid >= solv->strictrepopriorules && rid < solv->strictrepopriorules_end)
+    {
+      if (fromp)
+	*fromp = -r->p;
+      return SOLVER_RULE_STRICT_REPO_PRIORITY;
+    }
   if (rid >= solv->choicerules && rid < solv->choicerules_end)
     return SOLVER_RULE_CHOICE;
   if (rid >= solv->recommendsrules && rid < solv->recommendsrules_end)
@@ -3028,8 +3034,8 @@ solver_ruleclass(Solver *solv, Id rid)
     return SOLVER_RULE_CHOICE;
   if (rid >= solv->recommendsrules && rid < solv->recommendsrules_end)
     return SOLVER_RULE_RECOMMENDS;
-  if (rid >= solv->blackrules && rid < solv->blackrules_end)
-    return SOLVER_RULE_BLACK;
+  if (rid >= solv->strictrepopriorules && rid < solv->strictrepopriorules_end)
+    return SOLVER_RULE_STRICT_REPO_PRIORITY;
   if (rid >= solv->learntrules && rid < solv->nrules)
     return SOLVER_RULE_LEARNT;
   return SOLVER_RULE_UNKNOWN;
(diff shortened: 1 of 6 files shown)


src/rules.c
```diff
@@ -3028,8 +3034,8 @@ solver_ruleclass(Solver *solv, Id rid)
     return SOLVER_RULE_CHOICE;
   if (rid >= solv->recommendsrules && rid < solv->recommendsrules_end)
     return SOLVER_RULE_RECOMMENDS;
-  if (rid >= solv->blackrules && rid < solv->blackrules_end)
-    return SOLVER_RULE_BLACK;
```
adriendelsalle, line 3032: Note: the block was duplicate from L3025

adriendelsalle:
cc @wolfv

src/rules.c
```diff
@@ -4135,6 +4141,29 @@ solver_addrecommendsrules(Solver *solv)
...
+  Pool *pool = solv->pool;
+  Solvable *s;
+  Id sid, p, pp;
+
+  solv->strictrepopriorules = solv->nrules;
+  FOR_POOL_SOLVABLES(sid)
+  {
+	s = pool->solvables + sid;
+	int max_prio = 0;
+	FOR_PROVIDES(p, pp, s->name)
```
wolfv, line 4157: we could cache the max prio here since it's the same for each `s->name` id. @mlschroe do you have a suggestion how to make a map? maybe we can allocate an array with the size of `pool->ss.nstrings`? Or is there a better way?

wolfv: E.g.

int* cache = solv_calloc(pool->ss.nstrings, sizeof(int));

then we can use that null-initialized buffer to cache values:

if (cache[s->name] == 0)
{
For provides ... 

}
else
{
max_prio = cache[s->name] - 1;
}
...

mlschroe:
I would do it a bit different:

1. pass the "addedmap" to the function. It contains a map of all solvables relevant for the job. This limits the amount of work we do.
2. create a copy of the map
3. change the code so that it runs over the names, not the solvables. Thus we can reuse the max prio instead of calculating it again:

```
map_init_clone(&todo, addedmap);
FOR_POOL_SOLVABLES(p)
  {
    if (!MAPTST(&todo, p))
      continue;
    max_prio = s->repo->priority;
    s = pool->solvables + p;
    FOR_PROVIDES(p2, pp2, s->name)
      {
        Solvable *s2 = pool->solvables + p2;
        if (s->name != s2->name)
          continue;
        if (s2->repo->priority > max_prio)
          max_prio = s2->repo->priority;
      }
    FOR_PROVIDES(p2, pp2, s->name)
      {
        Solvable *s2 = pool->solvables + p2;
        if (s->name != s2->name || !MAPTST(&todo, p2))
          continue;
        MAPCLR(&todo, p2);
        if (s2->repo < max_prio)
          solver_addrule(solv, -p2, 0, 0);
      }
  }
```

mlschroe:
(The code above is completely untested)

src/rules.c
```diff
@@ -4135,6 +4141,29 @@ solver_addrecommendsrules(Solver *solv)
...
+{
+  Pool *pool = solv->pool;
+  Solvable *s;
+  Id sid, p, pp;
+
+  solv->strictrepopriorules = solv->nrules;
+  FOR_POOL_SOLVABLES(sid)
+  {
+	s = pool->solvables + sid;
+	int max_prio = 0;
```
wolfv, line 4156: if (s->repo == pool->installed) continue;

adriendelsalle: thx for catching this

src/rules.c
```diff
@@ -4135,6 +4141,54 @@ solver_addrecommendsrules(Solver *solv)
...
+	{
+		if (!MAPTST(&priomap, p))
+		  	continue;
+
+		s = pool->solvables + p;
+		max_prio = s->repo->priority;
+		FOR_PROVIDES(p2, pp2, s->name)
+		{
+		Solvable *s2 = pool->solvables + p2;
+		if (s->name != s2->name)
```
wolfv, line 4167: this probably never happens in the case of conda repodata, right?

src/rules.c
```diff
@@ -4135,6 +4141,54 @@ solver_addrecommendsrules(Solver *solv)
...
+		{
+		MAPCLR(&priomap, p2);
```
wolfv, line 4169: why mapclr here?

src/rules.c
mlschroe: It's good to do anyway as the strict repo prio code will probably be used by other distros as well.

src/rules.c
mlschroe: yes, that's wrong. you mustn't clear the bits in the first loop, as the second loop also does a MAPTST.

mlschroe: Oh wait, it only clears if the names are different. It's still wrong, though ;-)

Conan-Kudo:
I'm not sure "strict repo priority" is the correct wording for this behavior. This describes as "weak repo priority" to me, since this explicitly makes it so higher NVRs in another repo still wins (priorities normally make it so lower NVRs win anyway, at least in DNF and Zypper).

mlschroe:
The code doesn't check the version/release at all, so I don't understand your comment.

Conan-Kudo:
I was mostly going off #458.

src/rules.c
adriendelsalle: Sorry I think I got it, from `conda` packages names of providers are always the same so I take a wrong shortcut to clear them if names differ.

In the case where `s->name != s2->name` we don't clear `s2` but just continue to not prevent to add rules for `s2->name` later right?

adriendelsalle:
@mlschroe I updated that PR, is you have opportunity to take a look it would be great :)

wolfv:
@mlschroe we've been using this in mamba for a while and seems fine! Any objection to merging it?

SylvainCorlay:
+1 on getting this in. This appears to be fixing all the channel priority issues we've had with earlier versions.

Issue #461, opened 2021-07-18
chksum.raw() is broken on Python 3.10


ignatenkobrain:
```python
…
  File "/home/brain/Projects/upstream/rust2rpm/tools/fedora-helper.py", line 87, in get_binary_license
    cookie = chksum.raw()
  File "/usr/lib64/python3.10/site-packages/solv.py", line 1422, in raw
    return _solv.Chksum_raw(self)
SystemError: PY_SSIZE_T_CLEAN macro must be defined for '#' formats
```

`libsolv-0.7.17-5.fc35.x86_64`

There is some chance that it was already fixed but I did not have time to check, yet.

ignatenkobrain:
I guess it is coming from here

```c
%typemap(out,noblock=1,fragment="SWIG_FromCharPtrAndSize") BinaryBlob {
#if defined(SWIGPYTHON) && defined(PYTHON3)
  $result = $1.data ? Py_BuildValue("y#", $1.data, $1.len) : SWIG_Py_Void(); // ← HERE
#elif defined(SWIGTCL)
  Tcl_SetObjResult(interp, $1.data ? Tcl_NewByteArrayObj($1.data, $1.len) : NULL);
#else
  $result = SWIG_FromCharPtrAndSize($1.data, $1.len);
#if defined(SWIGPERL)
  argvi++;
#endif
#endif
}
```

PR #462, opened 2021-07-18
Fix compatibility with Python 3.10

ignatenkobrain:
Fixes: https://github.com/openSUSE/libsolv/issues/461
Signed-off-by: Igor Raits <igor.raits@gmail.com>

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -10,6 +10,12 @@
 %markfunc Pool "mark_Pool";
 #endif
 
+#ifdef SWIGPYTHON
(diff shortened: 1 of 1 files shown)


mlschroe:
Do we really need the cast in the call? I'm worried that Py_ssize_t is not available in python-2. (But I do agree about always defining PY_SSIZE_T_CLEAN).

mlschroe:
Oh wait, it's in a `#if defined(PYTHON3)` branch. Nevermind.

PR #469, opened 2021-08-09
Store conda packages signatures and signed metadata

adriendelsalle:
Description
---

- Store packages signatures
- Store packages extra metadata to allow later signature verification
  - those metadata are also signed
  - without storing metadata, we need to re-open and parse the `repodata.json` file
- Remove rounding of the timestamp, causing failing verification against original `repodata.json` file
  - another solution would be to have a signature process with that rounding step
  - it looks like introducing complexity
  - is this required for full features parity with `conda`? is there any case where 2 different builds differ from some milliseconds and should be considered equivalent?

cc @wolfv

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -149,6 +149,10 @@ parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn)
 
   handle = repo_add_solvable(pd->repo);
   s = pool_id2solvable(pool, handle);
+
+  Id extra_keys = pool_str2id(pool, "solvable:extra_keys", 1);
+  Id extra_values = pool_str2id(pool, "solvable:extra_values", 1);
+
   while (type > 0 && (type = jsonparser_parse(jp)) > 0 && type != JP_OBJECT_END)
     {
       if (type == JP_STRING && !strcmp(jp->key, "build"))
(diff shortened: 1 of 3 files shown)


mlschroe:
I don't think we should try to reconstruct the package info data from the solvable data. It is much easier to assemble the complete json block and store it in the solvable instead of the extra fields. It also has the advantage that we don't run into the number rounding issue.

There are different ways to move forward:

1) should package managers like mamba really use the libsolv implementation to parse the conda repo? You could as well create the libsolv packages in python. (I guess this will make the parsing a bit slower, though).
2) If we stick with libsolv's parser: do we want to support a c-only implementation without python? To verify a conda signature, the package data needs to be converted into normalized json. This normalization can either be done in python code or in libsolv. Which way do you prefer?

wolfv:
Hey @mlschroe, thanks for looking at this. 

Most of mamba's implementation is actually done in C++ (we also have a pure C++ client, called micromamba). So it's important to us that things work without Python.

We can use `nlohmann::json` (a C++ library) to normalize the JSON quite easily. So if we can get the un-normalized JSON from libsolv, that would probably be fine. 
We also have already implemented the code to re-assemble the proper JSON for the signature validation.

I think one downside of storing the entire JSON would be that the `.solv` files would get a bit larger, right?

Of course we could copy the repo_conda.c file or create our own implementation in mamba's C++ code for all that, too. So far, the code from libsolv has worked really well for us though :)

wolfv:
We would like to verify the signatures whenever a given package is installed.

mlschroe:
(Anyway, regarding the size of the solv files: we can put the signatures and the package info json into the "vertical" section of the solv file. This will make the code read it on demand and it will not be in memory all the time)

wolfv:
> Oh, so it's different to conda which verifies to signatures at repo parsing time?

Yes, exactly, because that is actually quite slow for large repositories ... 

> (Anyway, regarding the size of the solv files: we can put the signatures and the package info json into the "vertical" section of the solv file. This will make the code read it on demand and it will not be in memory all the time)

Sounds good! Although I always thought it would be cool to also try to use `.solv` files directly on the network and see if that works fine (and is faster), and that would only really make sense if the file is as small as possible :)

mlschroe:
I wouldn't advise to use them directly. The solv file parser does catch a lot of errors, but I wouldn't be surprised if there is a code path that can be be abused by feeding the parser an invalid file. It is meant as a cache file where we know that the content is good. No security guarantees....

wolfv:
Unfortunately, no. 
I just generated some signed repodata with the scripts in mamba's repo: https://github.com/mamba-org/mamba/tree/master/mamba/tests 

First run `generate_gpg_keys.sh` and then 

```
export KEY1=$(gpg --fingerprint "MAMBA1")
export KEY2=$(gpg --fingerprint "MAMBA2")
python reposerver.py -d repo/ --auth none --sign
```

this creates a `repo_signed` folder that contains some signed repodata and additional root key files.

```json
{
  "info": {
    "subdir": "noarch"
  },
  "packages": {
    "test-package-0.1-0.tar.bz2": {
      "build": "0",
      "build_number": 0,
      "depends": [],
      "license": "BSD",
      "license_family": "BSD",
      "md5": "2a8595f37faa2950e1b433acbe91d481",
      "name": "test-package",
      "noarch": "generic",
      "sha256": "b908ffce2d26d94c58c968abf286568d4bcf87d1cfe6c994958351724a6f6988",
      "size": 5719,
      "subdir": "noarch",
      "timestamp": 1613117294885,
      "version": "0.1"
    }
  },
  "packages.conda": {},
  "removed": [],
  "repodata_version": 1,
  "signatures": {
    "test-package-0.1-0.tar.bz2": {
      "f46b5a7caa43640744186564c098955147daa8bac4443887bc64d8bfee3d3569": {
        "signature": "0a50063539baf249970f1d08b07f00f544e2d87982826790e9ec6e80874ad90aec21a9607cf38bb58897163533c39cb4a4f1c741a7f8e9e4f67e2ff2087d2d00"
      }
    }
  }
}
```

adriendelsalle:
I did some benchmarks about storing all the metadata (necessary for package signature) or just the extras in an additional field [here](https://github.com/conda/conda/issues/10713#issuecomment-862350216)

adriendelsalle:
> Please try out commit [cc2da2e](https://github.com/openSUSE/libsolv/commit/cc2da2e789f651b2d0d55fe31c258426bf9e984d)

Why not continuing the job on this PR?
Don't hesitate to push on my branch

adriendelsalle:
I understand, for most of the problems listed we ensure reproducibility by normalization at serialization, including indentation and sorting.

mlschroe:
The json normalization sorts object keys, not arrays. And I was talking about whitespace in the dependency string, not whitespace in the json serialization.

adriendelsalle:
@mlschroe we started discussion with conda team about normalization, including much more that only key sorting.
Anyway, I'm closing this PR since you prefer to reboot the idea elsewhere

mlschroe:
Oh, if the conda team is willing to change things they could sign the sha512sum of the normalized json data. That wouldn't take up that much space.

Issue #471, opened 2021-08-11
Reading from Pool is not thread-safe when pool_alloctmpspace() is used


dmach:
Several functions use pool_alloctmpspace() to return a string allocated in Pool's temporary area.
Such strings must be processed and discarded immediately.
Everything works fine as long as the code runs in a single thread.

We're currently writing new dnfdaemon - a DBus interface that processes incoming requests in an infinite loop and turns them into asynchronous threads. We are aware of potential concurrency issues, designing the code to avoid them as much as possible.

Our use case is: load repositories into Pool and let users run multiple queries at a time.
We believe that pool_alloctmpspace() might become an issue when used in threads and that impacts all libsolv functions that use it, incl. pool_solvable2str() for example.

Locking doesn't seem to be a good idea, because a user would have to always unlock it explicitly after the value is consumed and it also decreases performance.
I'm wondering if `thread_local` annotation on `pool->tmpspace` couldn't be used to eliminate race conditions among threads.
Maybe also returning a newly allocated string might be an option, but I don't know the performance impact.
Or do you have any better idea?

ignatenkobrain:
I think more correct approach (which will also make it "safe" for multiple threads, assuming that you don't touch string pool, e.g. by creating new packages) is to mimic pool_solvable2str() yourself and use your temporary memory for those.

I think we could probably have 2 variations of these functions, so that caller can choose which one he wants (thread-safe or fast)…

They way current function is designed is to use as less memory as possible and to be as fast as possible.

dmach:
I don't think it's as easy as you suggested.
`pool->tmpspace` is used across the whole code incl. many lookups.

j-mracek:
@mlschroe I know multi threading is not an easy topic but I would like to ask you for your opinion. We are preparing a new DNF daemon and multi threading support in libsolv is a feature that is critical for the design. Do you think whether is it possible or feasible to make libsolv thread safe? Right now we see an issue with `pool->tmpspace` but there can be other workflows in libsolv that can not be compatible with multi thread usage. Anyway we will try our best to resolve the problem but at beginning we need to understand how big problem is in front of us.

mlschroe:
I don't think you can easily make libsolv multi-thread safe. You can do this for the tmpspace, but there's more to it. For example provider lookup caches the result is the whatprovides hash.

If we want to go this route we need to go through all the API functions and find out if they need locking. The original design was "a pool must only be accessed by one thread at a time", so there might be some ugly corner cases...

[...]

PR #472, opened 2021-08-17
Make reading from Pool more thread-safe

dmach:
Doesn't fix the problem completely, but mitigates it
so the pool can be accessed from tens of concurrent threads.

I tested pool_tmpjoin() and the performance is 40% worse (locking has a cost).
We need to test it in real-life scenarios, the overall impact should not be that bad.

Tries to fix https://github.com/openSUSE/libsolv/issues/471

### DIFF
--- a/doc/libsolv-pool.txt
+++ b/doc/libsolv-pool.txt
@@ -646,7 +646,10 @@ Utility functions
 
 Allocate space on the pool's temporary space area. This space has a limited
 lifetime, it will be automatically freed after a fixed amount (currently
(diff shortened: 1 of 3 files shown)


Conan-Kudo (approved):

j-mracek:
It looks like that the patch will not help much. First of all it does not solve the issue but it makes it less frequent. According to https://github.com/openSUSE/libsolv/issues/471#issuecomment-966930555 there are other problems in libsolv related to multi-thread safety.

PR #477, opened 2021-10-15
Add an ifdef allowing to rename conflicting Solvable::requires

lukash:
In C++20 requires is a new keyword, causing a conflict with the
Solvable::requires attribute. From the outside this can only be fixed
with an ugly preprocessor redefinition of the symbol around each libsolv
header include.

This commit adds an #ifdef allowing to rename the attribute in case
C++20 or higher is used, which can be easily defined in one place in the
libolv user's build system.

I am willing to resolve this in any other way if you have a better idea, as well as bikeshed on the new name for the attribute :slightly_smiling_face:.

### DIFF
--- a/src/solvable.h
+++ b/src/solvable.h
@@ -34,6 +34,19 @@ typedef struct s_Solvable {
   struct s_Repo *repo;		/* repo we belong to */
 
   /* dependencies are offsets into repo->idarraydata */
+  /* the ifdef resolves "requires" conflicting with a C++20 keyword */
+#ifdef LIBSOLV_SOLVABLE_PREPEND_DEP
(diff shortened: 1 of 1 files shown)


lukash:
I've changed the PR to rename all the dependency attributes with the `dep_` prefix.

Conan-Kudo (approved):

lukash:
@mlschroe ping

lukash:
@mlschroe gentle ping nr. 2. This is holding us back from switching to C++20 in our dnf 5 codebase and subsequent work.

mlschroe:
Ok ok, merged. Thanks!

lukash:
Thank you :slightly_smiling_face:

PR #478, opened 2021-10-21
Improvement: Add subdir extracted from repodata info if not found in individual package entries

johnhany97:
For context, please check https://github.com/mamba-org/mamba/pull/1227

To summarise the current situation:

The actual bug here seems to come from how different conda channel owners provide info about the packages.

So for example, conda-forge gives us both information about the subdir in the repodata.json `info` and in every package entry. Example:
```
{
  "info": {
    "subdir": "linux-64" // <--- Notice this
  },
  "packages": {
    "_anaconda_depends-2018.12-py27_0.tar.bz2": {
      "build": "py27_0",
      "build_number": 0,
      "depends": [ ... ],
      "license": "BSD",
      "md5": "199037865cc19536a1ae07b115e5a5c2",
      "name": "_anaconda_depends",
      "sha256": "cbaa2e02de8389a04f42ef98d92e11fb319a66dcd86834bdb434dd008525c593",
      "size": 5598,
      "subdir": "linux-64", // <--- Notice this too
      "timestamp": 1562173890182,
      "version": "2018.12"
    },
    ...
  },
  ...
}
```
and that works fine. But then when we use a channel like https://repo.continuum.io/pkgs/free (for which i have a repro of this behaviour in the mamba PR linked above), where their repodata.json has the subdir info only in the `info` block but not in every package entry, we fail.
```
{
  "info": {
    "arch": "x86_64",
    "platform": "linux",
    "subdir": "linux-64" // <--- Notice this, but the lack of another entry in the following package.
  },
  "packages": {
    "_libgcc_mutex-0.1-free.tar.bz2": {
      "build": "free",
      "build_number": 0,
      "date": "2019-07-01",
      "depends": [],
      "md5": "f7d7639c8485ae4701aeb6add534a774",
      "name": "_libgcc_mutex",
      "size": 3129,
      "timestamp": 1562011672620,
      "track_features": "free_channel_libgcc",
      "version": "0.1"
    },
  },
  ...
}
```

## Proposed solution

As we're parsing the repodata.json, we extract `subdir` out of the `info` block and then as we parse individual packages, we check if a parsed package entry has a missing `subdir` and if so, we override with whatever we managed to extract out of the repodata info block.

## Possible downsides

I believe the current approach works on the assumption we'll always parse the `info` block before we parse any of the `packages` (or `packages.conda`) sections. I need to figure out if there's a way to guarantee the order (or at least perform the parsing of the repodata twice, once for the subdir extraction, and another for package extraction).

### DIFF
--- a/ext/repo_conda.c
+++ b/ext/repo_conda.c
@@ -136,7 +136,7 @@ fn2data(struct parsedata *pd, const char *fn, Id *fntypep, int create)
 }
 
 static int
-parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn)
+parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn, char *repodata_subdir)
 {
   int type = JP_OBJECT;
   Pool *pool= pd->pool;
@@ -191,6 +191,9 @@ parse_package(struct parsedata *pd, struct solv_jsonparser *jp, char *kfn)
       else
 	type = jsonparser_skip(jp, type);
     }
+  /* some channels don't populate the individual packages' subdir property, so we add it from repodata info */
+  if (*subdir == 0)
+    subdir = repodata_subdir;
   if (fn || kfn)
     {
       repodata_set_location(data, handle, 0, subdir, fn ? fn : kfn);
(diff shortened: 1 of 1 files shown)


mlschroe:
We can't rely on rewinding the input stream. I implemented it with a different strategy in commit 9937d5f936cc70adeb5d06aea75ec7fdc92cd84c

PR #479, opened 2021-10-28
Ensure FIND_PACKAGE PythonInterp called before PythonLibs

markmcclain:
Upstream Cmake docs specify that PythonInterp should be called before
PythonLibs[1].  Inconsistency is possible when called with PythonLib first
in some environments.

[1] https://cmake.org/cmake/help/latest/module/FindPythonInterp.html


mlschroe:
Don't you also need to adjust the first lines of the cmake file, so that you get python2 if both python2 and python3 are to be built?

PR #484, opened 2022-01-06
Support loongarch64

zhangwenlong8911:
Signed-off-by: Zhang Wenlong <zhangwenlong@loongson.cn>


mlschroe:
There's no need for this as long as there are no architecture fallbacks. See commit aea953a42d99399938eb6c98a220310aa438c376

PR #488, opened 2022-02-08
Fix segfault on conflict resolution when using bindings

niner:
Solutionelement objects were created with a wrong solutionid (i.e. the
solution element id was used for the solutionid field). This led to invalid
array indexes when trying to read the extraflags which then escalated
into a segfault. Fix by setting solutionid correctly in the Solutionelement
constructor.


mlschroe:
Oh wow, how could that happen and stay unnoticed? Thanks!

(Btw, we need libsolv bindings for Raku ;-) )

Issue #496, opened 2022-04-27
Memory leak in tests detected by Address Sanitizer


Klaim:
# Context 

As part of investigation of some issues in [libmamba](https://github.com/mamba-org/mamba) which depends on `libsolv` I used Address Sanitizer (ASAN) and spotted a memory leak report coming from the lib. While there is probably mishandling on `libmamba`'s side, to be sure I attempted to build `libsolv` with ASAN and found a few tests leaking.

The following is on Ubuntu, using conda/mamba as environment, notably `gcc==10.3`

# How To Reproduce

0. clone `libsolve`, create a build directory
1. cmake invocation I used:
    ```
    cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG="-g -O0 -fno-omit-frame-pointer -fsanitize=address" -DLINK_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -DBUILD_TESTING=ON -DBUILD_SHARED=ON -S ../ -B ./
    ```
 2. Build and run the tests: `make && make test`
  
# Observed
 
The output:
```
[...]
96% tests passed, 1 tests failed out of 26

Total Test time (real) =   4.51 sec

The following tests FAILED:
          3 - choicerules (Failed)
Errors while running CTest
Output from these tests are in: /home/joel/work/libs/libsolv/build-debug/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```
In the log file we can find:
```
3/26 Test: choicerules
Command: "/home/joel/work/libs/libsolv/test/runtestcases.sh" "/home/joel/work/libs/libsolv/build-debug/tools/testsolv" "/home/joel/work/libs/libsolv/test/testcases/choicerules"
Directory: /home/joel/work/libs/libsolv/build-debug/test
"choicerules" start time: Apr 27 16:40 CEST
Output:
----------------------------------------------------------
testcase_read: missing feature 'complex_deps'
choice1b.t .......................................   Skipped

=================================================================
==255147==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1024 byte(s) in 1 object(s) allocated from:
    #0 0x7fdf5d685ea8 in __interceptor_malloc ../../../../libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7fdf5d5884f2 in solv_malloc /home/joel/work/libs/libsolv/src/util.c:38
[...]

PR #499, opened 2022-06-18
Fix memory leak when using testsolv to execute cases

banjiuqingshan:
*resultp will only keep the pointer of the last cycle, which will lead to memory leakage.
This solves the first memory leak problem in issue #496 "==255147==error..."


mlschroe:
Thanks!

PR #504, opened 2022-07-12
solv: add plaindir repo support

lnussel:


mlschroe:
You should ignore .patch.rpm and .delta.rpm files, like in tools/repo2solv.c's read_plaindir_repo() implementation.

lnussel:
updated

lnussel:
so any blockers that prevent merging?

Conan-Kudo:
Why do we want this?

mlschroe:
Thanks! (Although this is just example code, of course)

examples/solv/repoinfo_type_plaindir.c
```diff
@@ -0,0 +1,79 @@
...
+  if ((dp = opendir(cinfo->path)) == 0)
+    {
+      perror(cinfo->path);
+      return -1;
+    }
+  while ((de = readdir(dp)) != 0)
+    {
+      if (de->d_name[0] == 0 || de->d_name[0] == '.')
+        continue;
+      if (!endswith(de->d_name, ".rpm") || endswith(de->d_name, ".delta.rpm") || endswith(de->d_name, ".patch.rpm"))
```
Conan-Kudo, line 61: `*.drpm` is also a DeltaRPM extension...

mlschroe: Files not ending in `.rpm` are already ignored.

PR #505, opened 2022-08-15
Ensure duplinvolvedmap_all is reset when a solver is reused

jon-turney:
Otherwise, if solver is used with a SOLVER_DISTUPGRADE job, then reused
without one, this will cause solver_addduprules() to be called, even
though needduprules is 0, which will crash trying to dereference
solv->dupmap, which is NULL because solver_createdupmaps() hasn't been
called.


mlschroe:
Thanks a lot!

PR #509, opened 2022-09-12
Add support for computing hashes using OpenSSL

jan-kolarik:
Optional support for computing hashes using the OpenSSL library. 

The reason is that we would like to use a verified implementation of hash functions calculation on Fedora based systems. The OpenSSL library is installed by default there. 

This PR allows to turn on OpenSSL hash functions using the latest API which is supported since OpenSSL 1.1.0.


mlschroe:
But wouldn't that kill md5/sha1 if FIPS is turned on? (Just guessing...)

jan-kolarik:
I don't see an internal md5/sha1 usage except some test cases or am I missing something?

mlschroe:
Libsolv itself doesn't use it, but you don't know about the applications using libsolv (the C API or the bindings)

jan-kolarik:
> Libsolv itself doesn't use it, but you don't know about the applications using libsolv (the C API or the bindings)

Oh, right. I will look at that.

mlschroe:
(We could also use the rpmDigest* functions from librpmio.)

mlschroe:
Oh, and your pull request needs to do something about solv_chksum_create_clone(). I.e. you need to use EVP_MD_CTX_copy().

jan-kolarik:
> But wouldn't that kill md5/sha1 if FIPS is turned on? (Just guessing...)

I was thinking about it and isn't that like we are turning the FIPS on to get rid of those vulnerable functions? I'd expect anyone who want to use those should keep FIPS turned off.

> (We could also use the rpmDigest* functions from librpmio.)

I don't know how long this hash functions API is going to be maintained there, it's just a bridge to the actual crypto library. Also for non-RPM clients it would be probably unwanted dependency?

> Oh, and your pull request needs to do something about solv_chksum_create_clone(). I.e. you need to use EVP_MD_CTX_copy().

Thanks for the feedback. I updated the sources to handle cloning correctly and also to take care of the OpenSSL context cleanup.

Anyway, I discussed it with the DNF team and for now it is OK for us to have it as the downstream patch as we already have there the older version of this.

Issue #511, opened 2022-11-30
Memory leak caused by commit 1da9bef8 


banjiuqingshan:
=================================================================
==720210==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 0x7fe82b2e937f in __interceptor_calloc (/usr/lib64/libasan.so.6+0xaf37f)
    #1 0x7fe82b116baf in solv_calloc /root/rpmbuild/BUILD/libsolv-0.7.22/src/util.c:79
    #2 0x7fe82b06863a in map_init /root/rpmbuild/BUILD/libsolv-0.7.22/src/bitmap.c:24
    #3 0x7fe82b169d1f in solver_disablechoicerules /root/rpmbuild/BUILD/libsolv-0.7.22/src/rules.c:3577
    #4 0x7fe82b088cac in solver_check_unneeded_choicerules /root/rpmbuild/BUILD/libsolv-0.7.22/src/solver.c:2651
    #5 0x7fe82b08a34a in solver_run_sat /root/rpmbuild/BUILD/libsolv-0.7.22/src/solver.c:2845
    #6 0x7fe82b09ac00 in solver_solve /root/rpmbuild/BUILD/libsolv-0.7.22/src/solver.c:4267
    #7 0x403cf5 in main /root/rpmbuild/BUILD/libsolv-0.7.22/tools/testsolv.c:241
    #8 0x7fe82ae7726f  (/usr/lib64/libc.so.6+0x2d26f)

SUMMARY: AddressSanitizer: 1 byte(s) leaked in 1 allocation(s).

It has been determined that the memory allocated by map_init in solver_disablechoicerules failed to be released in time

banjiuqingshan:
Maybe this patch is not very appropriate, but it does solve the memory leak

PR #512, opened 2022-11-30
fix memory leak caused by commit 1da9bef8

banjiuqingshan:
fix memory leak caused by commit 1da9bef8.#511
the memory allocated by map_init in solver_disablechoicerules failed to be released in time.


mlschroe:
Can you please change this to `map_free(&m)`? Thanks!

PR #513, opened 2022-11-30
fix memory leak caused by commit 1da9bef8

banjiuqingshan:
the memory allocated by map_init in solver_disablechoicerules failed to be released in time.#511


mlschroe:
Thank you!

PR #516, opened 2022-12-19
Python: Remove usage of the deprecated distutils module

hroncok:
Python 3.12+ no longer has distutils.

See https://peps.python.org/pep-0632/


mlschroe:
Thank you!

PR #521, opened 2023-02-15
Fix conda headers

AntoinePrv:
Some fix to be able to include headers independently and without `extern "C"`.


mlschroe:
Thanks!

PR #530, opened 2023-05-03
Fix incomplete headers

AntoinePrv:


mlschroe:
I think an easier and cleaner fix would be to add a `#include "solver.h"` line. Would that work for your use case?

PR #532, opened 2023-06-19
Build python bindings for all python flavors

danigm:
This patch uses the python-rpm-macros to build the python bindings for all python flavors defined in the project.


Conan-Kudo:
This will break the build for non-SUSE distributions.

danigm:
> This will break the build for non-SUSE distributions.

Sorry, I'll try to update this change to make it compatible with non-SUSE

danigm:
I've updated the PR to do this with a flag `%singlespec_py3` that's just for modern SUSE versions.

PR #533, opened 2023-06-22
updateinfoxml: workaround memory problem when adding to two flexarrays

kontura:
When directly adding packages and modules to both `pd->handle` and `pd->collhandle` flexarrays right after each other `repodata_add_array` has to constantly move data in `data->attriddata`. It wastes a lot of memory.

For example running `updateinfoxml2solv` on fedora 38 updates repo [updateinfo](http://ftp.fi.muni.cz/pub/linux/fedora/linux/updates/38/Everything/x86_64/repodata/0e34d0301a70de30903a59f3008b7c154ac07977d6cc4b0a3da0e4fbbfa4fc65-updateinfo.xml.xz) uses ~30 MiB vs ~604 MiB (without patch) at usage peak.

### DIFF
--- a/ext/repo_updateinfoxml.c
+++ b/ext/repo_updateinfoxml.c
@@ -113,7 +113,8 @@ struct parsedata {
   Id pkghandle;
   struct solv_xmlparser xmlp;
   struct joindata jd;
-  Id collhandle;
+  Queue collectionpkgids;
+  Queue collectionmodids;
 };
 
 /*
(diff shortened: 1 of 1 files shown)


vpodzime:
Nice! :clap:

Conan-Kudo (approved):

mlschroe:
I would have preferred to fix the underlying problem with the add_array, by I don't see no easy way to do this. I did change your patch to just use one queue that stores both the type and the handle id.

Thanks a lot!

PR #551, opened 2024-02-09
bindings: Fix building with Python 3.13

jan-kolarik:
Use `PyObject_Call` instead of the deprecated `PyEval_CallObject` as this function is to be removed in Python 3.13.

See https://docs.python.org/3.13/whatsnew/3.13.html#id11.

Currently, libsolv fails to build with Python 3.13.0a1.

```
/builddir/build/BUILD/libsolv-0.7.25/redhat-linux-build/bindings/python/solv_python.c:4734:24: error: implicit declaration of function ‘PyEval_CallObject’; did you mean ‘PyObject_CallObject’? [-Werror=implicit-function-declaration]
 4734 |     PyObject *result = PyEval_CallObject((PyObject *)d, args);
      |                        ^~~~~~~~~~~~~~~~~
      |                        PyObject_CallObject
```

### DIFF
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -1729,7 +1729,7 @@ returnself(matchsolvable)
   SWIGINTERN int loadcallback(Pool *pool, Repodata *data, void *d) {
     XRepodata *xd = new_XRepodata(data->repo, data->repodataid);
     PyObject *args = Py_BuildValue("(O)", SWIG_NewPointerObj(SWIG_as_voidptr(xd), SWIGTYPE_p_XRepodata, SWIG_POINTER_OWN | 0));
-    PyObject *result = PyEval_CallObject((PyObject *)d, args);
+    PyObject *result = PyObject_Call((PyObject *)d, args, NULL);
     int ecode = 0;
     int vresult = 0;
     Py_DECREF(args);


Conan-Kudo (approved):

mlschroe:
We still need to be able to build against python-2, so there needs to be some if statements...

jan-kolarik:
> We still need to be able to build against python-2, so there needs to be some if statements...

I am not sure about all supported Python versions, but it seems the same signature is at least in Python 2.7: https://docs.python.org/2/c-api/object.html#c.PyObject_Call.

mlschroe:
Mmm, true. Maybe we should use `PyObject_CallObject()`?

mlschroe:
I found the following history in one of the python issues:

PyEval_CallObject() was added at Jan 12 1995 ([05360cd616ae](https://hg.python.org/lookup/05360cd616ae)). PyObject_CallObject(), PyObject_CallFunction() and PyObject_CallMethod() were added with [Include/abstract.h](https://github.com/python/cpython/blob/master/Include/abstract.h) at Jul 18 1995 ([d5398acafa2c](https://hg.python.org/lookup/d5398acafa2c)) and implemented in terms of PyEval_CallObject(). PyEval_CallObjectWithKeywords() was added few minutes later ([0261bf5b3819](https://hg.python.org/lookup/0261bf5b3819)). PyObject_Call() was added at Aug 02 2001 ([09df3254b49d](https://hg.python.org/lookup/09df3254b49d)) as a simple wrapper around tp_call. PyObject_CallFunction() and PyObject_CallMethod() were reimplemented in terms of PyObject_Call() at Aug 16 2002 ([255d1d3e66a3](https://hg.python.org/lookup/255d1d3e66a3)).

jan-kolarik:
> I found the following history in one of the python issues:
> 
> PyEval_CallObject() was added at Jan 12 1995 ([05360cd616ae](https://hg.python.org/lookup/05360cd616ae)). PyObject_CallObject(), PyObject_CallFunction() and PyObject_CallMethod() were added with [Include/abstract.h](https://github.com/python/cpython/blob/master/Include/abstract.h) at Jul 18 1995 ([d5398acafa2c](https://hg.python.org/lookup/d5398acafa2c)) and implemented in terms of PyEval_CallObject(). PyEval_CallObjectWithKeywords() was added few minutes later ([0261bf5b3819](https://hg.python.org/lookup/0261bf5b3819)). PyObject_Call() was added at Aug 02 2001 ([09df3254b49d](https://hg.python.org/lookup/09df3254b49d)) as a simple wrapper around tp_call. PyObject_CallFunction() and PyObject_CallMethod() were reimplemented in terms of PyObject_Call() at Aug 16 2002 ([255d1d3e66a3](https://hg.python.org/lookup/255d1d3e66a3)).

And in the last one there is the interesting message:
```
Squash a few calls to the hideously expensive PyObject_CallObject(o,a)
-- replace then with slightly faster PyObject_Call(o,a,NULL).  (The
difference is that the latter requires a to be a tuple; the former
allows other values and wraps them in a tuple if necessary; it
involves two more levels of C function calls to accomplish all that.)
```

jan-kolarik:
But I am no expert in any of these, I am completely fine with any fix to make the Python 3.13 builds succeed.

mlschroe:
ok, ok, I changed it to always use PyObject_Call (as it is available since python 2.2. and we always construct the args tupel). Thanks!

PR #557, opened 2024-03-26
Fix a couple small static analysis findings for uninitialized structs

dcantrell:
The memset() on the KeyValue is more explicit even though if you trace the code you will see it fills out the struct.  However, it's possible that not every struct member will be initialized and adding the memset() makes things more obvious and appeases the static analyzer.

The queue_init() appeared to just be missing.


mlschroe:
Thanks!

PR #559, opened 2024-04-22
Split libsolv-tools

dirkmueller:


mlschroe:
Wouldn't it be better to change the spec file so that it builds a library?

Conan-Kudo:
Please just use the shared library instead.

PR #560, opened 2024-04-23
package: Enable shared library for all platforms but SLE < 16

Conan-Kudo:
This simplifies things for interfacing with the library and ensures that when libsolv is updated that package managers get the correct functionality.

This also allows us to drop the hacks to deal with static libsolv for DNF in SUSE distributions.

### DIFF
--- a/package/libsolv.spec.in
+++ b/package/libsolv.spec.in
@@ -60,8 +60,13 @@
 %bcond_with ruby
 %bcond_with perl
 %endif
+%if 0%{?suse_version} && 0%{?suse_version} < 1600
 %bcond_without static
 %bcond_with shared
+%else
+%bcond_with static
+%bcond_without shared
+%endif
 %bcond_with zypp
 
 Name:           libsolv


Conan-Kudo:
`libsolvext` does not link properly as a static link library, that's why I have to watch what libsolvext is linked to and manually link them in to libdnf to make it work.

Conan-Kudo:
For reference:

* dnf5: https://build.opensuse.org/projects/system:packagemanager:dnf/packages/dnf5/files/dnf5-with-static-libsolvext.patch?expand=1
* libdnf: https://build.opensuse.org/projects/system:packagemanager:dnf/packages/libdnf/files/libdnf-0.72.0-with-static-libsolvext.patch?expand=1

I would really like to drop these, but that requires a dynamic link library build for libsolv and libsolvext.

Conan-Kudo:
Additionally, statically linking libsolv has the major downside that when libsolv updates are pushed, there is no guarantee that the enhancements will affect the package managers using it, particularly in SLE/Leap, where automatic reverse dependency rebuilds do not happen.

Conan-Kudo:
Can we _please_ land this? It's painful to have to manually deal with this mess.

mlschroe:
We're now building both the static and the dynamic libraries on openSUSE/SLES16

PR #563, opened 2024-05-13
Report unsupported compression in solv_xfopen() with errno

ppisar:
If libsolv was built without Zstandard support and "primary.xml.zst" was passed solv_xfopen(), solv_xfopen() returned 0 without setting errno. A calling application could not distinguish an unsupported compression format from other I/O errors.

This patch improves this situation by setting errno variable to an appropriate value.


mlschroe:
Is ENOTSUP available everywhere? I.e. on BSD and windows (MSVC)?

(You can also use solv_xfopen_iscompressed() to figure out if the compression is supported: it returns 0 if the file is not compressed, 1 if it is compressed and the compression is supported, or -1 if the compression is not supported.)

ppisar:
I have no idea. Do people process YUM repositories on BSD or Windows?

ppisar:
ENOTSUP is documented in POSIX 2017 <https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html>.

Conan-Kudo:
> I have no idea. Do people process YUM repositories on BSD or Windows?

Yes. But also this is not YUM-only.

ppisar:
Ok. I will add a #ifdef guard.

ppisar:
Now the code should compile even if ENOTSUP is not available.

mlschroe:
I think if it's in POSIX we don't need the guard. If somebody complains that it's missing for some specific OS, we can add a
```
#ifndef ENOTSUP
#define ENOTSUP <something_else>
#endif
```
at the beginning...

ppisar:
I removed the guards.

mlschroe:
Thanks!

PR #565, opened 2024-06-07
Add a Conflicts in libsolv-tools-base

bmwiedemann:
Add a Conflicts in `libsolv-tools-base`
so this does not get installed in parallel with older versions of `libsolv-tools`


mlschroe:
I think we should use %version instead of that fixed version. We need the version to be in sync.

PR #566, opened 2024-06-19
Specify link dependencies for static build

bzeller:
For static linking we need to specify the link dependencies via the pkg-config and FindLibSolv.cmake files as well, so statically linked application can refer to them.


Conan-Kudo:
Hopefully this will let me get rid of my hack in the DNF build scripts to deal with static libsolv, though I still don't know why we just don't ship libsolv as a shared library (everyone else does...).

PR #567, opened 2024-06-20
Move findutils requires into libsolv-tools-base

dcermak:
repo2solv requires find, but it is now part of tools-base

### DIFF
--- a/package/libsolv.spec.in
+++ b/package/libsolv.spec.in
@@ -172,6 +172,7 @@ Summary:        Utilities used by libzypp to manage .solv files
 Group:          System/Management
 Provides:       libsolv-tools:%{_bindir}/repo2solv
 Conflicts:      libsolv-tools < %{version}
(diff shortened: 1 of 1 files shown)


bmwiedemann (approved):

mlandres:
```
[master]SUSE:tools (0)> repo2solv /Local/Repo >/dev/null 
/usr/bin/find: No such file or directory
find: exit status 1
[master]SUSE:tools (0)> 
                                     ^ $?
```
And maybe `repo2solv` should return not-zero. Currently it creates an empty repo cache without indicating the isue.

dirkmueller:
#569  is an alternative that avoids reintroducing this dependency, thereby keeping the installsize of zypper small.

mlschroe:
I now implemented the file tree walk in libsolv, so there is no need to relay on an external tool anymore.

PR #569, opened 2024-07-02
Use bash for installing rpms from plain directory

dirkmueller:
Also give an exit code in case globbing failed.


ignatenkobrain:
but this is not equal to the previous code?

dirkmueller:
> but this is not equal to the previous code?

I have tested it for my usecases, and it behaves similarly. what issue do you see ?

Conan-Kudo:
Doesn't this mean that openSUSE has to guarantee bash in minimal environments now?

dirkmueller:
> Doesn't this mean that openSUSE has to guarantee bash in minimal environments now?

we already do have this guarantee.

PR #575, opened 2024-12-21
[cmake] Require CMake 3.10

gruenich:
More recent versions warn that support for CMake < 3.10 is deprecated and will be eventually removed.


Conan-Kudo (approved):

mlschroe:
Merged with pull request #581

PR #576, opened 2024-12-21
Remove file for travis CI

gruenich:
This should be no longer used for a long time


mlschroe:
OK. Lets see if someone complains...

PR #592, opened 2025-07-01
Add pool_dep_fulfilled_in_map function

kontura:
This can be used to determine dependency closure on a given set of packages (even with complex dependencies).

It is basically `solver_dep_fulfilled(...)` but without `Solver`.

I have created this when looking for a solution to https://github.com/rpm-software-management/dnf-plugins-core/issues/549

### DIFF
--- a/doc/libsolv-pool.txt
+++ b/doc/libsolv-pool.txt
@@ -562,6 +562,10 @@ by the ``evr'' parts must overlap.
 Like pool_match_dep, but the provider is the "self-provides" dependency
 of the Solvable _s_, i.e. the dependency ``s->name = s->evr''.
 
+        int pool_dep_fulfilled_in_map(Pool *pool, const Map *map, Id dep);
+
+Returns ``1'' if the dependency _dep_ is provided by at least one package
+from _map_, otherwise ``0'' is returned.
 
 Whatprovides Index
 ------------------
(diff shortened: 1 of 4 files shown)


mlschroe:
Hi, just as a heads up: I'm fine with adding this but I don't like the name of the function.

kontura:
> Hi, just as a heads up: I'm fine with adding this but I don't like the name of the function.

Would just `pool_dep_fulfilled` be better?

mlschroe:
The _map prefix is fine (though I would drop the `_in`, because we don't have it in the other functions that have a `installedmap` argument). I'll merge the pull request now, I'll do the rename later.

mlschroe:
I think it's either `pool_dep_satisfied_map` or `pool_dep_fulfilled_map`. Any preferences?

kontura:
> I think it's either `pool_dep_satisfied_map` or `pool_dep_fulfilled_map`. Any preferences?

Personally I would prefer fulfilled but only because it is already used in the solver version.

Either way, thank you.

mlschroe:
Looking at other dependency management systems I seems like `satisfied` seems to be a bit more common. So I tend to go with `pool_dep_satisfied_map` (or `pool_satisfieddep_map` because we already have `pool_whatmatchesdep` and `pool_whatcontainsdep`).

PR #598, opened 2025-09-08
Adds WITH_OPENSSL build option

jonathanspw:
If it is ON, OpenSSL will be used instead of internal implementation of computing hashes (MD5, SHA1, SHA224, SHA256, SHA384, SHA512).

Original patch by Jaroslav Rohel <jrohel@redhat.com>.  I've done rebasing work to get it compatible with the current master branch.

This has been carried downstream in CentOS/RHEL for a while but we see no reason why it cannot be upstreamed.

https://gitlab.com/redhat/centos-stream/rpms/libsolv/-/merge_requests/28


Conan-Kudo:
Why is "patch.patch" checked in?

PR #600, opened 2025-10-22
Fix qsort_r preprocessor for musl and FreeBSD

goodspeed34:
The original mentioned qsort_r signature difference now only exists in DragonFly BSD & MacOS. However, the preprocessor also broke the compliation on musl+linux and FreeBSD, leading the compilation error on buildroot.

musl: https://git.musl-libc.org/cgit/musl/commit/?id=b76f37fd5625d038141b52184956fb4b7838e9a5
freebsd, dragonfly, macos: QSORT(3)

FreeBSD and musl use the same GNU-like signature.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -154,26 +154,14 @@ solv_setcloexec(int fd, int state)
   #endif
 }
 
-/* bsd's qsort_r has different arguments, so we define our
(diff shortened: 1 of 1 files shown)


mlschroe:
Looks good, thanks!

PR #603, opened 2025-12-18
Remove superfluous cast of NULL pointer

cmeerw:
NetBSD (and OpenBSD) actually use off_t instead of fpos_t here, so the cast introduces an unnecessary diagnostic when compiling on NetBSD.

NetBSD man page: [funopen.3](https://man.netbsd.org/NetBSD-10.1/funopen.3)
OpenBSD man page: [funopen.3](https://man.openbsd.org/funopen.3)
FreeBSD man page: [funopen.3](https://man.freebsd.org/cgi/man.cgi?query=funopen&sektion=3&manpath=FreeBSD+15.0-RELEASE&arch=default&format=html)


mlschroe:
thanks!

